PR #128, opened 2016-03-17
Initial smoke test for Kubernetes.

ewiseblatt:
This is only create/delete server group.

Kubernetes support for the observer looks awkward because kubectl has
awkward commands and doesnt fully support json. But it is sufficient for
now for this test. Might need to take a different approach when writing
future tests.

@lwander
This can wait for tomorrow. I'm not going to do any more today.


ewiseblatt:
github is confused by this branch and thinks there are conflicts.
I created pull/145 using a git diff master with this branch and it is happy so am closing this one.

PR #136, opened 2016-04-06
Refactor PathPredicate to handle multiple results and improve traceability

ewiseblatt:
A discussion of the refactoring is in the following document:
https://docs.google.com/document/d/1ggXbkyH7VZkKB1GPo8rtz80UQWH8sEp9XPxL22j3

The gist is this can handle embedded lists and provide explicit traceability
both to the paths of objects used and justification why other paths were
truncated. There are some more simplifications here, though the data model
is a bit richer so can involve more traversal and indirection. This wont
affect reporting significantly, other than making richer data available
and easier to navigate (by removing ambiguity and providing more
explicit traceability).

@lwander This is essentially the same as before but going o a refactoring branch rather than directly into the master branch


lwander:
Prior discussion was here: https://github.com/google/citest/pull/134

Looks good to me still.

Also, can you open up that doc?

ewiseblatt:
Link may have changed because it is different now. I updated the git comment
https://docs.google.com/document/d/1ggXbkyH7VZkKB1GPo8rtz80UQWH8sEp9XPxL22j3U_8

PR #137, opened 2016-04-06
Refactored remaining predicate modules.

ewiseblatt:
Moved these into json_predicate package, leaving contract verification in
json_contract. All the predicate modules should be working, but not yet
contracts or anything above it.

The main change here is that CardinalityPredicate is no longer based on Mapping,
rather it is based on PathPredicate.  I think this is more intuitive and does
not need to distinguish between mapped and unmapped constraints. The reporting
is better and more complete as well. The CardinalityPredicate needs the
HasPathPredicateResult hack so verifiers will be able to pull the values it
collected. This isnt ideal so will likely be revisted later. However for the
time being, doing it this way minimizes propagation of API changes.


ewiseblatt:
@lwander This is the second. The binary/cardinality tests should have been moved from before, but arent showing up that way on github.

citest/json_predicate/binary_predicate.py
```diff
@@ -428,8 +458,19 @@ def __call__(self, value):
     '==', lambda a, b: a == b, operand_type=list)
 LIST_NE = StandardBinaryPredicateFactory(
     '!=', lambda a, b: a != b, operand_type=list)
+
+def lists_equivalent(a, b):
```
lwander: The semantics of this are different from `set(a) == set(b)`, is that OK?

citest/json_predicate/logic_predicate.py
```diff
@@ -112,6 +118,9 @@ def __init__(self, pred):
   def __str__(self):
     return 'NOT ({0})'.format(self.__pred)
```
lwander: You can force the `__str__` representation by changing this to `{0!s}`. I think that's the default, but it might be clearer since you're forcing the `__repr__` representation below.

citest/json_predicate/binary_predicate.py
ewiseblatt: I cannot use sets because the lists often contain dictionaries which are not hashable so cannot convert here. That's why I wrote my own function.

lwander: That makes sense. I meant `set([1, 2, 2]) == set([1, 2])`, while `!lists_equivalent([1, 2, 2], [1, 2])`.

lwander:
LGTM

citest/json_predicate/logic_predicate.py
ewiseblatt: Isnt {0} well understood as string conversion?.
I'm going to change the __repr__ to produce structured JSON instead of
adhoc strings so that it can be reformatted to be readable. The strings get
very nasty as they compose to higher and higher levels. But am not going to
change it until later because of other priorities.

citest/json_predicate/binary_predicate.py
ewiseblatt: Good point. I didnt consider that. This way is more correct. Probably
should have been a bag before because multiple "identical" values suggests
something different is going on.

PR #138, opened 2016-04-07
Refactored observers, verifiers, and contracts.

ewiseblatt:
@lwander This is step 3, updating the contract and observers to be aware of the changes to the predicates. The main change here is that the verifier is now more based on PathPredicate than MapPredicate. Even so, since the results are only used for reporting and are self describing, there isnt further down-stream impact

### DIFF
--- a/citest/json_contract/contract.py
+++ b/citest/json_contract/contract.py
@@ -28,7 +28,7 @@ import time
 
 from ..base import JournalLogger
 from ..base import JsonSnapshotable
-from . import predicate
+from ..json_predicate import predicate
 from . import observer as ob
 from . import observation_verifier as ov
 
@@ -82,6 +82,11 @@ class ContractClauseVerifyResult(predicate.PredicateResult):
         self.__clause.title,
         self.__verify_results.enumerated_summary_message)
 
+  def __repr__(self):
+    return '{0!r} clause={1!r} verify_results={2!r}'.format(
+        super(ContractClauseVerifyResult, self).__repr__(),
+        self.__clause, self.__verify_results)
+
   def export_to_json_snapshot(self, snapshot, entity):
     """Implements JsonSnapshotable interface."""
     builder = snapshot.edge_builder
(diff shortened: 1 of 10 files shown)


citest/json_contract/contract.py
```diff
@@ -82,6 +82,11 @@ def __str__(self):
         self.__clause.title,
         self.__verify_results.enumerated_summary_message)
 
+  def __repr__(self):
+    return '{0} clause={1!r} verify_results={2!r}'.format(
+        super(ContractClauseVerifyResult, self).__repr__(),
```
lwander: Why not use `{0!r}` here?

citest/json_contract/observation_verifier.py
```diff
@@ -51,7 +51,6 @@ def __init__(self, observation):
     self.__valid_obj_set = []  # Cannot be a set because it is unhashable.
     self.__good_results = []
     self.__bad_results = []
-    self.__all_results = []
```
lwander: Why is this no longer needed?

citest/json_contract/observation_verifier.py
```diff
@@ -64,35 +63,47 @@ def __add_valid_object_constraint(self, entry):
...
+    path_pred_result = has_path_pred_result.path_predicate_result
+    good = path_pred_result.valid_candidates
+    attempts = [map_predicate.ObjectResultMapAttempt(
+        elem.path_value.value, elem.result) for elem in good]
+    self.__good_results.extend(attempts)
+    for entry in attempts:
+      self.__add_valid_object_constraint(entry)
+
+    failures = path_pred_result.invalid_candidates
+    self.__bad_results.extend([map_predicate.ObjectResultMapAttempt(elem.path_value.value, elem.result) for elem in failures])
```
lwander: Line > 80 chars

citest/json_contract/observation_verifier.py
```diff
@@ -64,35 +63,47 @@ def __add_valid_object_constraint(self, entry):
...
+
+    if not has_path_pred_result:
```
lwander: Why is not having a result considered a failure?

citest/json_contract/observation_verifier.py
ewiseblatt: Trying to simplify. all was superfluous. I thought it would be convienent to look at as a whole, but I dont recall ever using it. In practice you just want to see what was bad (if expecting everything to be good) and/or what was good (if expecting some but not all to be good)

lwander:
LGTM

PR #139, opened 2016-04-07
Updated remaining modules affected by refactoring to complete it.

ewiseblatt:
@lwander This is the final refactoring step, mostly updating the package name in the consumers. Since most high level usage is at the contract level, not much is affected.

There is a subtle problem here. The observers want to expand out lists but sometimes contracts are written against the list as a whole (contains an element that...). This is easily fixed by specifying the DONT_ENUMERATE_TERMINAL path, however it is a subtle issue that is going to grow into an annoyance. I'll need to address this, but have higher priority concerns in other projects so need to come back to it later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,51 +12,69 @@ if appropriate.
 A test case might look something like this:
 
 ```
-  def create_load_balancer(self):
-    load_balancer_name = self._bindings['TEST_APP_COMPONENT_NAME']
+  def create_google_load_balancer(self):
+    bindings = self._bindings
+    load_balancer_name = bindings['TEST_APP_COMPONENT_NAME']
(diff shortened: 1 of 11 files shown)


lwander:
LGTM

The issue you're describing: could you use a layer of indirection where the contract is applied against the list object, and the elements of the list are expanded separately?

ewiseblatt:
that's what this is doing. The contract is against the list (DONT_EXPAND_TERMINAL) then the list is expanded inside the filter (CONTAINS ...). I think I might be able to do this in the API either through different methods or another parameter. I'm going to let that bake for a while and stick with being explicit in the path expression itself for now.

PR #140, opened 2016-04-07
Added retry to observer in google kato deregister lb test.

jtk54:
Added retry to observer in google kato deregister lb test. Combats test flakiness due to timing.


jtk54:
@ewiseblatt PTAL

PR #141, opened 2016-04-07
Moved deregister lb retry to clause creation.

jtk54:
Moved deregister lb observer retry to clause creation.


jtk54:
@ewiseblatt PTAL.

PR #142, opened 2016-04-07
Refactored json_contract package

ewiseblatt:
Refactored PathPredicate to handle multiple results and improve traceability.

Moved predicate related definitions/operations to json_predicate
json_contract is observations, contracts, and their verification.

The gist is this can handle embedded lists and provide explicit traceability
both to the paths of objects used and justification why other paths were
truncated. There are some more simplifications here, though the data model
is a bit richer so can involve more traversal and indirection. This wont
affect reporting significantly, other than making richer data available
and easier to navigate (by removing ambiguity and providing more
explicit traceability).

@lwander This is a merge of the previous feature branch rebased into master


lwander:
LGTM

PR #144, opened 2016-04-11
Tweaked some snapshot details.

ewiseblatt:
@lwander

### DIFF
--- a/citest/json_predicate/path_predicate_result.py
+++ b/citest/json_predicate/path_predicate_result.py
@@ -228,14 +228,25 @@ class PathPredicateResult(predicate.PredicateResult, HasPathPredicateResult):
   def export_to_json_snapshot(self, snapshot, entity):
     """Implements JsonSnapshotable interface."""
     snapshot.edge_builder.make_mechanism(entity, 'Predicate', self.__pred)
+    # Separate out just the path values from the full justification (below)
+    # to make it easy to get at the list of values found, which is often
+    # of particular interest.
     snapshot.edge_builder.make_output(
         entity, 'Path Values', self.__path_values)
-    snapshot.edge_builder.make_output(
-        entity, 'Pruned Paths', self.__path_failures)
-    snapshot.edge_builder.make_output(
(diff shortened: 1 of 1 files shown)


citest/json_predicate/path_predicate_result.py
```diff
@@ -230,12 +230,15 @@ def export_to_json_snapshot(self, snapshot, entity):
...
-        entity, 'Pruned Paths', self.__path_failures)
-    snapshot.edge_builder.make_output(
-        entity, 'Rejected Values', self.__invalid_candidates)
-    snapshot.edge_builder.make_output(
-        entity, 'Value Justification', self.__valid_candidates)
+    if self.__path_failures:
+      snapshot.edge_builder.make_output(
+          entity, 'Pruned Paths', self.__path_failures)
+    if self.__invalid_candidates:
+        snapshot.edge_builder.make_output(
```
lwander: Indentation is off

citest/json_predicate/path_predicate_result.py
```diff
@@ -230,12 +230,15 @@ def export_to_json_snapshot(self, snapshot, entity):
...
+          entity, 'Rejected Values', self.__invalid_candidates)
+    if self.__valid_candidates:
+      snapshot.edge_builder.make_output(
+          entity, 'Value Justification', self.__valid_candidates)
```
lwander: What is a Value Justification?

lwander:
LGTM

citest/json_predicate/path_predicate_result.py
ewiseblatt: Added comment (and changed to Value Justifications).

FYI the snapshot breaks out (duplicates) just the values because that's commonly what you want to see. However the snapshot provides the full traceability of where those values came from and why they were chosen. That is the justification. The "Rejected Values" are the same thing but for the values that didnt match the criteria. To complete the picture "Pruned Paths" are where a value could not be found at all (e.g. point out where an internal attribute might not exist because it had a different name, was collapsed, or had another intermediate node).

PR #145, opened 2016-04-11
Initial smoke test for Kubernetes.

ewiseblatt:
This replaces the earlier branch because, for some reason, it doesnt think it is mergable. I hope this is.
It is [should be] exactly the same but is recreated from a new branch off master using a diff against it.

@lwander


lwander:
Does this test pass as is?

ewiseblatt:
it does.

lwander:
Sweet, let's merge it then.

PR #147, opened 2016-04-13
Expose max_wait_secs per operation (and make it a float)

ewiseblatt:
@ttomsu This change would have let you change the per-operation timeout yesterday


ttomsu:
LGTM

PR #152, opened 2016-05-25
Small typo fixes

ewiseblatt:
@lwander

### DIFF
--- a/citest/service_testing/agent_test_case.py
+++ b/citest/service_testing/agent_test_case.py
@@ -59,6 +59,7 @@ class OperationContractExecutionAttempt(JsonSnapshotable):
 
   @property
   def completed(self):
+    """The time the attempt completed, or None if it is still running."""
     return self.__stop is not None
 
   @property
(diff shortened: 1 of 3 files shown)


citest/service_testing/agent_test_case.py
```diff
@@ -540,7 +542,7 @@ def run_test_case(self, test_case, timeout_ok=False,
         self.logger.debug('Exception was at:\n%s', traceback.format_exc())
       except BaseException as unexpected:
         self.logger.error(
-            'Unexpected error {0}\nHandling original exception {1}',
+            'Unexpected error %s\nHandling original exception %s',
```
lwander: Why use `%s` over `{0}`?

lwander:
LGTM

citest/service_testing/agent_test_case.py
ewiseblatt: I'm using positional parameters to the logger. I think {0} is only for format strings. Or at least I was getting a lint error here.

PR #153, opened 2016-05-25
Added HTTP put support.

ewiseblatt:
@lwander 
This also removes out a non-standard use of trace parameter that was causing a lint error. Eventually I will remove all trace of "trace".


lwander:
LGTM

PR #154, opened 2016-05-25
Added HttpObservationFailureVerifier

ewiseblatt:
@lwander 
The "_error_comment_or_none" in the base class interface is hacky but I'd like to wait until later to clean it up. This is an enabling CL to add some new spinnaker tests in another CL.


lwander:
Testing for failure is a good idea.

LGTM

PR #156, opened 2016-06-02
Make test check against /health instead of /

ttomsu:
@lwander @jtk54 PTAL

@ewiseblatt FYI - A recent change to Gate put a redirect on / to Deck, which citest does not bring up.

### DIFF
--- a/citest/gcp_testing/gce_util.py
+++ b/citest/gcp_testing/gce_util.py
@@ -287,7 +287,7 @@ def establish_network_connectivity(gcloud, instance, target_port):
     atexit.register(_ProcessKiller(pid).safe_kill)
 
   logger.debug('Confirming tunnel is working')
-  url = 'http://localhost:%d' % local_port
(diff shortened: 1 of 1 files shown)


lwander:
LGTM

the rational is that /health is a more sane endpoint for checking application status that /.

citest/gcp_testing/gce_util.py
```diff
@@ -287,7 +287,7 @@ def establish_network_connectivity(gcloud, instance, target_port):
     atexit.register(_ProcessKiller(pid).safe_kill)
 
   logger.debug('Confirming tunnel is working')
-  url = 'http://localhost:%d' % local_port
+  url = 'http://localhost:%d/health' % local_port
```
ewiseblatt: This isnt right. This module is in citest, not spinnaker, so is intended to be generic. To add this, you need to add a "basepath" attribute and have spinnaker add the /health path.

Since this is already merged, I'll look into fixing it.

PR #158, opened 2016-06-06
Added GoogleCloudStorage agent

ewiseblatt:
@lwander

### DIFF
--- a/citest/gcp_testing/gcs_contract.py
+++ b/citest/gcp_testing/gcs_contract.py
@@ -1,0 +1,203 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Provides a means for specifying and verifying expectations of GCE state."""
+
+# Standard python modules.
+import json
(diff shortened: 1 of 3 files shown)


citest/gcp_testing/gcs_contract.py
```diff
@@ -0,0 +1,193 @@
...
+      disjunction_builder = jc.ObservationVerifierBuilder(
+          'Retrieve {0} bucket={1} path={2} or 404'.format(type, bucket, path))
+      disjunction_builder.append_verifier(error_verifier)
+
+      retrieve_builder = jc.ValueObservationVerifierBuilder(
+          'Retrieve {0} bucket={1} path={2}'.format(type, bucket, path),
+          strict=self.__strict)
+      disjunction_builder.append_verifier_builder(
+          retrieve_builder, new_term=True)
+      self.verifier_builder.append_verifier_builder(
```
lwander: Can you explain why you nest the verifiers in the order that you do? This is a bit confusing to read.

citest/gcp_testing/gsutil_agent.py
```diff
@@ -0,0 +1,88 @@
...
+    result = self.run(args)
+    if not result.ok:
+      return result
+
+    found = []
+    for item in result.output.split('\n'):
+      marker = item.find('#')
+      if marker < 0:
+        info = {'name': item}
+      else:
```
lwander: Is this code path ever hit if `with_versions` is false?

ewiseblatt: no, why?

citest/gcp_testing/gcs_contract.py
ewiseblatt: Added a comment. It would be nice to have internal API support for this. The data model is straight forward, but to get the simple caller API you have to pull this slight of hand.

citest/gcp_testing/gsutil_agent.py
lwander: Just wondering after playing with API since you didn't make it explicit.

lwander:
LGTM

PR #160, opened 2016-06-07
Fix MIG/RC terminology usage

lwander:
@ewiseblatt 

I'm going to enable this test in the nightly builds next.

### DIFF
--- a/spinnaker/spinnaker_system/kube_smoke_test.py
+++ b/spinnaker/spinnaker_system/kube_smoke_test.py
@@ -59,8 +59,8 @@ class KubeSmokeTestScenario(sk.SpinnakerTestScenario):
   This scenario defines the different test operations.
   We're going to:
     Create a Spinnaker Application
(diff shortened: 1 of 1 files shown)


spinnaker/spinnaker_system/kube_smoke_test.py
```diff
@@ -230,7 +230,7 @@ def create_server_group(self):
             }],
             'stack': bindings['TEST_STACK'],
```
ewiseblatt: You might want to test both ways now that we can create a service.
(actually it was fixed when I did the original checkin but forgot to add it)

ewiseblatt:
I'll let you do another PR to add the test since there are probably others you'd add as well.

spinnaker/spinnaker_system/kube_smoke_test.py
lwander: Ah, hah, yes, whoops.

PR #163, opened 2016-06-09
Kubernetes find image test

lwander:
@ewiseblatt PTAL


spinnaker/spinnaker_system/kube_find_image_test.py
```diff
@@ -0,0 +1,519 @@
...
+  """
+
+  @classmethod
+  def new_agent(cls, bindings):
+    """Implements citest.service_testing.AgentTestScenario.new_agent."""
+    agent = gate.new_agent(bindings)
+    agent.default_max_wait_secs = 70
+    return agent
+
+  @classmethod
```
ewiseblatt: If you arent adding custom arguments you dont need this.
I've put it in as a placeholder expecting to do it later.

spinnaker/spinnaker_system/kube_find_image_test.py
```diff
@@ -0,0 +1,519 @@
...
+      agent: [GateAgent] The agent for invoking the test operations on Gate.
+    """
+    super(KubeFindImageTestScenario, self).__init__(bindings, agent)
+    bindings = self.bindings
+
+    optional_stack = ('-{0}'.format(bindings['TEST_STACK'])
+                      if bindings['TEST_STACK']
+                      else '')
+
+    # No detail because name length is restricted too much to afford one!
```
ewiseblatt: This pattern is common in a lot of tests. I wonder if gate module should provide a shortcut for it like it does to create applications.

spinnaker/spinnaker_system/kube_find_image_test.py
```diff
@@ -0,0 +1,519 @@
...
+    # because it scopes the context of our activities.
+    # pylint: disable=invalid-name
+    self.TEST_APP = bindings['TEST_APP']
+
+    self.pipeline_id = None
+
+    # We will deploy two images. One with a tag that we want to find,
+    # and another that we don't want to find.
+    self.__image_registry = 'gcr.io'
+    self.__image_repository = 'kubernetes-spinnaker/test-image'
```
ewiseblatt: Where does the name come from?
As a rule of thumb, names are unique to the test so that multiple instances can run concurrently (or leak things without affecting other tests). There is a unique  test_id property on the scenario you can decorate with.

spinnaker/spinnaker_system/kube_find_image_test.py
ewiseblatt: If you are scraping configuration stuff from the runtime server, then you might need commandline overrides for non-GCE deployments where you cannot ssh into the machine to inspect the configuration as we do. At least there is no implementation for it at this time.

spinnaker/spinnaker_system/kube_find_image_test.py
lwander: It's a static image in on a public repository in GCR - it's not modified by the test so there is no worry about parallel runs using this image. It doesn't show up in the config anywhere because tags are resolved at runtime.

spinnaker/spinnaker_system/kube_find_image_test.py
lwander: A gate module to generate app-stack-detail names? Good idea. I'll do that in the next PR.

spinnaker/spinnaker_system/kube_find_image_test.py
lwander: Ah, I thought I needed this to make sure I get the regular arguments. Nothing is parsed from the config at this point since all settings are left at their default value. I could change that in the future, but that would make the config in spinnaker/spinnaker/install/first_google_boot.sh a bit more complicated than it needs to be.

spinnaker/spinnaker_system/kube_find_image_test.py
lwander: I think it should even live in it's own module "frigga" which we can add under the spinnaker specific code, since Gate isn't really in charge of generating those names in the first place.

lwander: I added a commit to address the naming for this test. I'll extend it to the others in a separate PR.

spinnaker/spinnaker_system/kube_find_image_test.py
```diff
@@ -311,9 +309,9 @@ def make_find_image_stage(self, requisiteStages=[], **kwargs):
 
   def make_deploy_stage(self, imageSource=None, requisiteStages=[], **kwargs):
     bindings = self.bindings
-    cluster = '{app}-{stack}'.format(
+    cluster = str(frigga.Cluster(
```
ewiseblatt: Where is this used?

spinnaker/spinnaker_testing/frigga.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Implementation of Netflix's Frigga naming convention"""
+
```
ewiseblatt: Why have these as classes rather than functions that produce the names? You dont use the instances at all. You only stringify them.

lwander: I wasn't sure what the preferred way was to do this. I can switch it to a function per name type.

spinnaker/spinnaker_system/kube_find_image_test.py
lwander: [Here](https://github.com/google/citest/pull/163/files/b350935defc797984c81eee23c36297bb0f561f9..5223587f354f162a4e557d68523d5243df6956b1#diff-a163732ccac0eba8a465d7b7af9c2b4bL338) and [here](https://github.com/google/citest/pull/163/files/b350935defc797984c81eee23c36297bb0f561f9..5223587f354f162a4e557d68523d5243df6956b1#diff-a163732ccac0eba8a465d7b7af9c2b4bL341)

lwander:
I think all your comments are addressed, @ewiseblatt

lwander:
I renamed this to "smoke test", since this is a basic, fundamental workflow with a cursory look at a number of Kubernetes/Spinnaker features.

lwander:
I'm ready to merge when you are

spinnaker/spinnaker_system/kube_smoke_test.py
```diff
@@ -243,22 +270,130 @@ def create_server_group(self):
...
+      'requisiteStageRefIds': requisiteStages,
+      'refId': 'FINDIMAGE',
+      'type': 'findImage',
+      'name': 'Find Valid Image',
+      'cloudProviderType': 'kubernetes', 
+      'namespaces': ['default'], 
+      'cloudProvider': 'kubernetes',
+      'selectionStrategy': 'NEWEST',
+      'onlyEnabled': True,
+      'credentials': 'my-kubernetes-account',
```
ewiseblatt: Credentials should be obtained from the deployed configuration.  See any of the other tests for how to do this.

spinnaker/spinnaker_system/kube_smoke_test.py
```diff
@@ -330,7 +496,7 @@ def main():
 
   defaults = {
       'TEST_STACK': 'tst',
-      'TEST_APP': 'kubsmok' + KubeSmokeTestScenario.DEFAULT_TEST_ID
+      'TEST_APP': 'kubfimg' + KubeSmokeTestScenario.DEFAULT_TEST_ID
```
ewiseblatt: why fimg? I liked smok better since more obvious to track back resources to this test.

spinnaker/spinnaker_testing/frigga.py
ewiseblatt: looks good. thanks.

ewiseblatt:
Fix the credential thing, rename the TEST_APP and you can merge.

spinnaker/spinnaker_system/kube_smoke_test.py
lwander: Whoops, artifact of renaming. Will fix.

PR #165, opened 2016-06-13
Added GoogleSession logout.

jtk54:
Added GoogleSession logout. @ewiseblatt FYI.

### DIFF
--- a/citest/gcp_testing/google_session.py
+++ b/citest/gcp_testing/google_session.py
@@ -32,6 +32,7 @@ class GoogleSession(object):
     req = urllib2.Request(url=login_url)
     cookie_jar.add_cookie_header(req)
     resp = urllib2.urlopen(req)
+    session.logout()
   """
 
   def __init__(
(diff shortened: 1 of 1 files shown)


citest/gcp_testing/google_session.py
```diff
@@ -73,7 +74,8 @@ def __init__(
   def logout(self):
     """Log out of the current authenticated Google session.
     """
-    raise NotImplementedError()
+    self.__session.get("https://www.google.com/accounts/Logout")
+    self.__session.cookies.clear()
```
ewiseblatt: Is clearing the cookies ok here, or does that change behavior to an unauthenticated call? Could this result in future calls being unauthenticated (and passing with some different behavior for anonymous clients) vs failing because the provided cookie is invalid?
Just curious. Otherwise LGTM

jtk54: This makes future calls the the scope of this session anonymous/unauthenticated, since `cookies` would be empty. This would cause a redirect if you make a subsequent call to try to re-access the same service, since it no longer recognizes your session.

PR #166, opened 2016-06-13
Handle session logouts gracefully.

jtk54:
Handle session logouts gracefully. @ewiseblatt PTAL. @ttomsu FYI.

### DIFF
--- a/citest/gcp_testing/google_session.py
+++ b/citest/gcp_testing/google_session.py
@@ -27,7 +27,11 @@ class GoogleSession(object):
     # login_url should forward to 'accounts.google.com/ServiceLogin' page
     # via an OAuth2 or SAML redirect.
     login_url = "http://<gate_host>/credentials"
-    session = GoogleSession(login_url, "user@domain.net", "secret_pass_word")
+    logout_url = "http://<gate_host>/auth/logout"
+    session = GoogleSession(login_url,
+                           "user@domain.net",
+                           "secret_pass_word",
+                           logout_url=logout_url)
     cookie_jar = session.cookies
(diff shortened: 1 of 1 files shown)


citest/gcp_testing/google_session.py
```diff
@@ -69,13 +77,17 @@ def __init__(
         approval_dict[input['name']] = input['value']
     approval_dict['submit_access'] = 'true'
     self.__session.post(approval_form['action'], data=approval_dict)
```
ewiseblatt: I'm not diggin' the logout_url here or the flow in logout(). At least not now. It is hooking things together without being clear with what this means or is for or why the caller doesnt. On its own, this is authenticating a session. It isnt intrinsically on behalf of another party. So to logout for another party is just weird, and adding a registry of people to notify is overkill without more justification.

ewiseblatt: discussed offline. ok.

citest/gcp_testing/google_session.py
```diff
@@ -87,4 +99,6 @@ def cookies(self):
       session as a CookieJar object.
       CookieJar documentation: https://docs.python.org/2.7/library/cookielib.html#cookiejar-and-filecookiejar-objects
     """
+    if not self.__session.cookies:
```
ewiseblatt: is None? Might not matter since you'll always have non-empty cookies otherwise?

jtk54: Done, it's theoretically possible to have zero cookies after login so we'll explicitly check it this way.

PR #168, opened 2016-06-14
fixed package setup

ewiseblatt:
@jtk54


README.md
```diff
@@ -8,6 +8,10 @@ where the core framework accomodates for asynchronous calls and even retries
 if appropriate.
 
 
+# Setup
+pip install -r requirements.txt
```
jtk54: Should we add something about setting up a `virtualenv`, or are we not worrying about dev environments in this scenario?

jtk54: Addressed offline.

jtk54:
LGTM

PR #171, opened 2016-06-15
Use python API to Google Cloud Storage rather than gsutil

ewiseblatt:
@lwander, @jtk54


citest/gcp_testing/gcp_api.py
```diff
@@ -0,0 +1,76 @@
...
+def build_authenticated_service(name, version, scopes, credentials_path):
+  """Create an authenticated service client instance.
+
+  Args:
+    name: [String] The name of the service.
+    version: [String] The service API version.
+    scopes: [String] The credentials OAuth2 scopes.
+    credentials_path: [String] Path to JSON file containing GCP secrets.
+  Returns:
+    A Resource object wiht methods for interacting with the service.
```
jtk54: with

citest/gcp_testing/gcp_api.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+"""Helper functions for creating authenticated GCP service clients."""
+
+import logging
+import httplib2
+
+from .. import json_contract as jc
+
+from apiclient.discovery import build as discovery_build
```
jtk54: I'd do `import apiclient.discovery as discovery` and then call `discovery.build` to be clearer here.

citest/gcp_testing/google_cloud_storage_agent.py
```diff
@@ -0,0 +1,126 @@
...
+FULL_SCOPE = 'https://www.googleapis.com/auth/devstorage.full_control'
+
+
+class GoogleCloudStorageAgent(base_agent.BaseAgent):
+  """Agent that interacts with Google Cloud Storage service."""
+
+  def __init__(self, json_credential_path, scope):
+    """Constructs agent.
+
+    Args:
```
jtk54: Add `scope` as an arg here.

citest/gcp_testing/google_cloud_storage_agent.py
```diff
@@ -0,0 +1,126 @@
...
+          service account.
+    """
+    super(GoogleCloudStorageAgent, self).__init__()
+    self.__service = build_authenticated_service(
+        'storage', 'v1', scope, json_credential_path)
+
+  def inspect(self, bucket, path=None, generation=None):
+    """Get metadata for a bucket or object in the bucket.
+
+    Args:
```
jtk54: Add `generation` as an arg here.

citest/gcp_testing/google_cloud_storage_agent.py
```diff
@@ -0,0 +1,126 @@
...
+                                             generation=generation)
+    response = request.execute()
+    self.logger.info('%s', response)
+    return response
+
+  def list(self, bucket, path, with_versions, fields=None):
+    """List the contents of the path in the specified bucket.
+
+    Args:
+      bucket: [string] The name of the bucket to list.
```
jtk54: In the previous function docstrings above you had [String] instead of [string].

citest/gcp_testing/google_cloud_storage_agent.py
```diff
@@ -0,0 +1,126 @@
...
+    self.logger.info('Listing bucket=%s prefix=%s versions=%s',
+                     bucket, path, with_versions)
+    request = self.__service.objects().list(bucket=bucket,
+                                            prefix=path, fields=fields,
+                                            versions=with_versions)
+    all_objects = []
+    while request:
+      response = request.execute()
+      all_objects.extend(response.get('items', []))
+      request = self.__service.objects().list_next(request, response)
```
jtk54: Is there any way to batch these calls?

spinnaker/spinnaker_system/google_front50_test.py
```diff
@@ -97,12 +102,13 @@ def create_app(self):
     # file.
     gcs_builder = gcp.GoogleCloudStorageContractBuilder(self.gcs_observer)
     (gcs_builder.new_clause_builder('Created Google Cloud Storage File')
-     .list(self.BUCKET_ROOT, 'applications')
+     .list(self.BUCKET, os.path.join(self.BASE_PATH, 'applications'))
```
jtk54: `os.path.join` will join the args based on the path separator that the current os thinks is the path separator. This is probably fine for our use cases, but on Windows this might break.

citest/gcp_testing/google_cloud_storage_agent.py
ewiseblatt: Only by increasing the result limit. This is chaining together pages, which is inherently serial.

jtk54: Ok, makes sense.

jtk54:
LGTM after resolving merge conflict and rebasing. I think it's `setup.py` that is in conflict.

PR #174, opened 2016-06-17
Better support for expecting errors from GCP APIs

ewiseblatt:
@jtk54

### DIFF
--- a/citest/gcp_testing/gcp_api.py
+++ b/citest/gcp_testing/gcp_api.py
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 """Helper functions for creating authenticated GCP service clients."""
 
+
 import logging
+import re
+
 import httplib2
 
 from apiclient import discovery
@@ -23,6 +25,7 @@ from googleapiclient.errors import HttpError
 from oauth2client.service_account import ServiceAccountCredentials
 
 from .. import json_contract as jc
+from .. import json_predicate as jp
 
 
 PLATFORM_READ_ONLY_SCOPE = (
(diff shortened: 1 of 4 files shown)


ewiseblatt:
@jtk54

citest/gcp_testing/gcp_api.py
```diff
@@ -52,26 +55,84 @@ def build_authenticated_service(name, version, scopes, credentials_path):
...
-        and error.match_regex(self.__error_regex)):
-      return 'Error matches {0}'.format(self.__error_regex)
+  def __call__(self, observation):
+    """Check if observation contains the expected HTTP error.
+
+    Args:
+      observation: [Observation] The observation
+    """
+    message = None
+    error = None
```
jtk54: This value is not overridden anywhere else after it's initialized. I would also rename it if it is used, since you shadow the same name in the following loop.

citest/gcp_testing/gcp_api.py
```diff
@@ -52,26 +55,84 @@ def build_authenticated_service(name, version, scopes, credentials_path):
...
+          # Not the expected error
+          message = None
+          continue
+
+      # Error matched.
+      break
+
+    # Error does not match.
+    if message:
+      error_result = jc.ObservationFailedError(valid=True, failures=[error])
```
jtk54: from L89, `error` will always be `None`. I'm not sure what the intended behavior is here.

citest/gcp_testing/gcp_api.py
ewiseblatt: This is the same error as the loop index. It is initialized to None so it is defined if observation.errors is empty.

citest/service_testing/http_observer.py
```diff
@@ -147,16 +147,33 @@ def __init__(self, title, agent, retryable_for_secs=0, strict=False):
     self.__agent = agent
     self.__strict = strict
 
-  def get_url_path(self, path):
+  def get_url_path(self, path, expect_http_error=None):
     """Perform the observation using HTTP GET on a path.
 
     Args:
       path [string]: The path relative to the agent's baseUrl.
+      expect_http_error: [int] If not None then we expect this HTTP error
```
jtk54: `expect_http_error` -> `expected_http_status_code`.

citest/gcp_testing/gcp_api.py
jtk54: Ok, I was just skeptical of the scoping rules. This could be a little clearer if the outer `error` was renamed and explicitly set in the loop.

jtk54:
LGTM after comments.

ewiseblatt:
I refactored this to clarify (and simplify) the data model. PTAL

citest/gcp_testing/gcp_api.py
```diff
@@ -52,26 +55,155 @@ def build_authenticated_service(name, version, scopes, credentials_path):
...
+class HttpErrorPredicate(jp.ValuePredicate):
+  """Determine if an object matches an expected HttpError."""
+
+  @property
+  def http_code(self):
+    """The expected HTTP code or None."""
+    return self.__http_code
+
+  @property
+  def contains_regex(self):
```
jtk54: Maybe change this to `content_regex`?

citest/gcp_testing/gcp_api.py
```diff
@@ -52,26 +55,155 @@ def build_authenticated_service(name, version, scopes, credentials_path):
...
+
+  def __call__(self, value):
+    if not isinstance(value, HttpError):
+      return jp.PredicateResult(
+          valid=False,
+          comment='{0} != HttpType'.format(value.__class__.__name__))
+
+    message = ['HttpError']
+    sep = 'with'  # Separator between confirmation phrases.
+    if self.__http_code:
```
jtk54: `if self.__http_code is not None:` since it's explicitly set to `None` in the case where it doesn't exist.

citest/gcp_testing/gcp_api.py
```diff
@@ -52,26 +55,155 @@ def build_authenticated_service(name, version, scopes, credentials_path):
...
+      if self.__http_code != value.resp.status:
+        return HttpErrorPredicateResult(
+            False, value,
+            comment='HTTP Status={0} != {1}'.format(value.resp.status,
+                                                    self.__http_code))
+
+      message.append('{0} HTTP Status={1}'.format(sep, self.__http_code))
+      sep = 'and'
+
+    if self.__contains_regex:
```
jtk54: Same as L115.

jtk54:
I like the refactoring, makes the logic clearer and easier to read. LGTM after comments and rebase.

PR #176, opened 2016-06-28
Added script for GCP quota checking in jenkins builds.

jtk54:
Added script for GCP quota checking in jenkins builds. @lwander PTAL. @ewiseblatt FYI.


spinnaker/spinnaker_testing/gcp_quota_check.py
```diff
@@ -0,0 +1,72 @@
...
+                      type=float,
+                      help='Percentage of use your quota should be under.')
+  parser.add_argument('--credentials_path',
+                      help='Path to your GCP service account key.')
+  args = parser.parse_args()
+  if (not args.quota_threshold
+      or not args.gcp_project
+      or not args.credentials_path):
+    logging.error('Invalid values for input args: must not be None.')
+    raise ValueError('Invalid values for input args: must not be None.')
```
lwander: Given that this is in citest, I recommend hooking into Eric's existing logging/verification tools

spinnaker/spinnaker_testing/gcp_quota_check.py
```diff
@@ -0,0 +1,72 @@
...
+  if args.quota_threshold > 100.0 or args.quota_threshold < 0.0:
+    logging.error(
+      'Invalid value for arg quota_threshold: must be between 0.0 and 100.0.')
+    raise ValueError(
+      'Invalid value for arg quota_threshold: must be between 0.0 and 100.0.')
+  return args
+
+def main():
```
lwander: This should probably be formatted like the other tests.

spinnaker/spinnaker_testing/gcp_quota_check.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
lwander: This strikes me as a weird place to write this test. The rest of this directory is full of classes that act as utilities for testing Spinnaker subsystems.

lwander:
Jake and I discussed this out-of-band, and the conclusion was that this isn't the best way to do this.

1. it prevents tests not dependent on GCP from running
2. it doesn't use all of the built in journaling and logging that Eric has written

The solution will be to rewrite this to allow each test that will create GCP infrastructure to call this before running a scenario to see if enough resources exist.

citest/service_testing/cli_agent.py
```diff
@@ -156,7 +156,9 @@ def _args_to_full_commandline(self, args):
     This is an internal method used to ensure that the arguments are ultimately
     consistent if needed, regardless of how the agent is invoked.
     """
-    return [self.__program] + args
+    # The filter() here removes empty strings, which cause some cli
+    # commands to fail.
+    return filter(None, [self.__program] + args)
```
lwander: When were empty strings being added? I feel like this should fail if someone is incorrectly providing bad args to the `cli_agent`

jtk54: There are valid `gcloud` commands where `name` is empty/null, such as `gcloud compute project-info describe`, so I tried to call it both with `None` and an empty string, neither of which worked. I pushed another change that removes this filter and fixes the `gcloud` command construction.

spinnaker/spinnaker_system/google_smoke_test.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015 Google Inc. All Rights Reserved.
+# Copyright 2016 Google Inc. All Rights Reserved.
```
lwander: I think we try to leave this at the date the file was created.

spinnaker/spinnaker_testing/spinnaker_test_scenario.py
```diff
@@ -378,6 +378,8 @@ def __update_bindings_with_subsystem_configuration(self, agent):
     it is testing.
     """
     # pylint: disable=bad-indentation
+    if agent is None:
```
lwander: Semantically, what does this mean/how does this occur?

lwander:
Nice, LGTM.

For posterity: This PR now contains a revised version of Jake's original approach which instead of creating a separate test (as the original PR did), creates a test stage to run before any Google test that requires GCP quota. This way each test that can fail due to quota constraints, will fail for the appropriate reason.

spinnaker/spinnaker_testing/spinnaker_test_scenario.py
jtk54: This case occurs if we don't have/want an agent that is making calls to Spinnaker (as is the case with the quota test).

PR #178, opened 2016-07-08
Added service account to quota tests commands.

jtk54:
Added service account to quota tests commands. They used to depend on the active gcloud service account, this PR makes them much more robust. @lwander @danielpeach PTAL. @ewiseblatt FYI.


danielpeach:
LGTM

lwander:
LG2M2

PR #179, opened 2016-07-16
Enable decoupling of AgentTestScenario and AgentTestCase.

ewiseblatt:
Added hook to base runner to initialize argument parser (so can inject
scenario specific arguments) and provided a means for the TestRunner to
store shared data across TestCase instances. This still isnt ideal, but
is a simple change that will remove part of the messy coupling and
simplify bootstrap.

@lwander or @jtk54

### DIFF
--- a/citest/base/test_runner.py
+++ b/citest/base/test_runner.py
@@ -99,6 +99,7 @@ class TestRunner(object):
   """
 
   __global_runner = None
+  __singleton_scenario_instances = {}
 
   @property
   def options(self):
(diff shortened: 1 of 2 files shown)


lwander:
How are data written to the shared store?

LGTM

ewiseblatt:
The "shared store" is just an object. The base test class acts as a
singleton factory (using a class variable). The class doesnt need to be a
singleton, but the base class manages the instance to use so that all
instances of the derived test class (i.e. all tests) will use the same
attribute value thus sharing the same data object across all instances.
See https://github.com/google/citest/pull/180 for how this is used (the PR
contains a few additional changes). In short, rather than having a
"scenario" attribute added by a ScenarioTestRunner where the test injects
the scenario instance to use and there is a bit of a dance to coordinate
the AgentTestCase with the Scenario, they can just for the shared_data with
the desired scenario class from the base TestCase. That gets rid of the
need for ScenarioTestRunner and setup coordination.

PR #180, opened 2016-07-16
Use shared data tests

ewiseblatt:
This is a followup to the earlier PR that enabled decoupling of AgentTestScenario by removing the use of the AgentScenarioTestRunner from all the existing tests and, since it is no longer used, removing it entirely.

This also removes the explicit default account bindings for the quota test. It doesnt make sense to be hardcoding those since it is deployment dependent. When run as a developer, it will use the default gcloud credentials (the ones provided wouldnt make sense). Our jenkins job already supplies the credentials via command-line argument, so this change shouldnt break the build.

@lwander or @jtk54


ewiseblatt:
This includes the earlier PR because that wasnt approved yet. I'm intentionally making these separate but can squash and submit it as one if you prefer to review that way.

spinnaker/spinnaker_system/google_quota_test.py
```diff
@@ -24,6 +24,7 @@
```
ewiseblatt: I'm going to submit this change as a separate PR. Baking in default account credentials doesnt make sense here. The account info should have been passed as command line arguments by the caller (e.g. jenkins task) or left as the default, which is the typical use case for individual developers running from the command line.

ewiseblatt:
rebased to earlier PR

lwander:
LGTM

PR #182, opened 2016-07-18
Fix quota test

ewiseblatt:
@lwander or @jtk54


citest/gcp_testing/quota_predicate.py
```diff
@@ -0,0 +1,170 @@
...
+            PathValueResult(valid=False, pred=EQUIVALENT(metric),
+                            source=value, target_path='metric',
+                            path_value=PathValue('', available.keys())))
+        continue
+
+      found = available[metric]
+
+      # TODO(ewiseblatt): 20160715
+      # Need some way to express a DIFFERENCE of two paths to say
+      # that we want to evaluate ('limit' - 'usage') then report these values.
```
jtk54: Maybe this is too general, but I think something like a custom `LambdaPredicate` or something similar that takes a function and runs it on the input JSON would be useful. We could then extend that to basic +, -, *, / operations.

spinnaker/spinnaker_system/google_server_group_test.py
```diff
@@ -10,11 +10,25 @@
...
 import spinnaker_testing.gate as gate
-import google_quota_test as quota
 import citest.base
 
 
-class GoogleServerGroupTestScenario(quota.GoogleQuotaTestScenario):
+class GoogleServerGroupTestScenario(sk.SpinnakerTestScenario):
+
+  MINIMUM_PROJECT_QUOTA = {
+      'INSTANCE_TEMPLATES': 1,
```
jtk54: Since we reuse these, maybe a quota constants file would be helpful to have rather than poking around with gcloud to find the quota keys.

spinnaker/spinnaker_system/google_smoke_test.py
```diff
@@ -68,6 +67,21 @@ class GoogleSmokeTestScenario(quota.GoogleQuotaTestScenario):
     Delete each of the above (in reverse order)
   """
 
+  MINIMUM_PROJECT_QUOTA = {
+      'INSTANCE_TEMPLATES': 1,
```
jtk54: See reuse comment above.

jtk54:
LGTM after comments, I like the refactor.

PR #183, opened 2016-07-21
Added path transforms

ewiseblatt:
@jtk54 

This is an initial stab at providing support to interpret implied values for the other CL.
I expect to change this later so that transforms return ValueResult instead of values, but that wont affect how they are consumed (only their unit testing and reporting) so putting that off.


jtk54:
LGTM

PR #186, opened 2016-07-26
Introduce and use GcpComputeAgent over GCloudAgent

ewiseblatt:
@jtk54


citest/reporting/html_renderer.py
```diff
@@ -176,7 +176,7 @@ def __tr_for_html_info(self, label_html, info,
     formatter.push_level()
     data_indent = formatter.line_indent
     formatter.pop_level(3)
-
+
```
jtk54: Remove whitespace.

jtk54:
LGTM, much cleaner than gcloud.

PR #189, opened 2016-08-04
Remove AgentTestCase's dependency on TestRunner.

ewiseblatt:
Dependency was only for storing objects in the journal.
Since it is already using JournalLogger, and store only forwards to the
global runner's logger, it asks the JournalLogger to store the object instead.

Also added test for AgentTestCase.

@jtk54


jtk54:
LGTM after resolving conflicts.

PR #191, opened 2016-08-04
Make JsonSnapshotable a bit more generic.

ewiseblatt:
Renamed JsonSnapshotable to JsonSnapshotableEntity meaning that the JSON
representation is an entity (composite value). Added JsonSnapshotable base
class with a method that returns a JSON value that can be added into a snapshot
representation.

The difference is that objects can be made "JsonSnapshotable" that return
simple values rather than composite values. This will allow wrapper objects
(for edge values) that can return simple values without having to expose the
wrapper to the snapshot.

This is also a bit simpler API, even just for JsonSnapshotableEntity.

Also, renamed the snapshot _for_data methods as _for_object because
they refer to "entities" (which are composite objects) and not simple
data objects (e.g. wrapped values). Simple objects arent added to snapshots,
though are used for edge attributes describing relationships between objects
in a snapshot.

@lwander

### DIFF
--- a/citest/aws_testing/aws_agent.py
+++ b/citest/aws_testing/aws_agent.py
@@ -57,7 +57,7 @@ class AwsAgent(st.CliAgent):
     self.logger = logging.getLogger(__name__)
 
   def export_to_json_snapshot(self, snapshot, entity):
-    """Implements JsonSnapshotable interface."""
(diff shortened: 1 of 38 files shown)


lwander:
LGTM

PR #194, opened 2016-08-05
Added gcp_contract test

ewiseblatt:
refactored from gcp_agent_test and borrowed from gcloud_contract_test


ewiseblatt:
@lwander

PR #195, opened 2016-08-05
Add base.ExecutionContext parameter to contract verification

ewiseblatt:
@lwander review, @jtk54 FYI

This is the first PR that the earlier refactoring setup. It is what I described to you earlier, but I added an "eval" method which is how it is consumed -- you run your values through the context and it transforms them into the actual values. So if the values are literal, they remain literal. But if they are functions then they get evaluated. The assumption is that functions have a signature taking one parameter, which is an ExecutionContext.

As I noted offline, the internal vs snapshotable attributes in the context is speculative assuming that I was going to log the context but not snapshot everything since they are redundant with other things in the snapshot. It might be a bit cumbersome so am open to other suggestions.


citest/base/execution_context.py
```diff
@@ -0,0 +1,151 @@
...
+      return value(self)
+    else:
+      return value
+
+  def eval_list(self, values):
+    if values is None:
+      return None
+    return [elem(self) if callable(elem) else elem for elem in values]
+
+  def eval_dict(self, values):
```
lwander: Why have `eval_dict` as well as `eval` which seems to operate on dictionaries, but doesn't call `eval_dict`?.

citest/base/execution_context.py
```diff
@@ -0,0 +1,151 @@
...
+      return {key: self.eval(data) for key, data in value.items()}
+    elif callable(value):
+      return value(self)
+    else:
+      return value
+
+  def eval_list(self, values):
+    if values is None:
+      return None
+    return [elem(self) if callable(elem) else elem for elem in values]
```
lwander: How is the execution state passed into the evaluator?

lwander:
Nice, this is going to be really helpful, LGTM

citest/base/execution_context.py
ewiseblatt: Removed eval_dict and eval_list; they turned out not to be needed or used.

citest/base/execution_context.py
ewiseblatt: Removed eval_list.
The execution is the context (self) and is the evaluator. The element is the callback (or value at #136). The execution state is the parameter into the callable that is providing the value.

PR #197, opened 2016-08-08
Added context to predicate evaluation

ewiseblatt:
Renamed CloneableInContext to CloneableWithNewSource to avoid
overloading the term "context". The context in journals and evaluation
are still overloaded but expect them to become explicitly related in
the future.

@lwander this is the last of the CLs for now. This completes the feature for runtime value injection from an execution context.

### DIFF
--- a/citest/json_predicate/binary_predicate.py
+++ b/citest/json_predicate/binary_predicate.py
@@ -48,19 +48,45 @@ class BinaryPredicate(predicate.ValuePredicate):
     """The fixed operand argument."""
     return self.__operand
 
+  @property
+  def operand_type(self):
+    """The expected type of the operand."""
+    return self.__operand_type
+
+  def eval_context_operand(self, context):
+    """Determine the operand type for the given evaluation context."""
+    operand = context.eval(self.__operand)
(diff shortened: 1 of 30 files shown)


citest/json_predicate/binary_predicate.py
```diff
@@ -139,20 +150,23 @@ class DictSubsetPredicate(BinaryPredicate):
   """Implements binary predicate comparison predicates against dict values."""
 
   def __init__(self, operand, **kwargs):
-    if not isinstance(operand, dict):
+    if not (isinstance(operand, dict) or callable(operand)):
       raise TypeError(
           '{0} is not a dict: {1!r}'.format(operand.__class__, operand))
```
lwander: Shouldn't the type error reflect the possible failed `callable(operand)` predicate?

lwander:
LGTM

citest/json_predicate/binary_predicate.py
ewiseblatt: callable(operand) determines whether operand is callable or not. If the operand is callable, then we defer the check until later when the value is resolved.

If it is callable, the disjunction is true so the guard is false.
If it is not callable, then the disjunction is just as before.

PR #200, opened 2016-08-10
Fixed execution context

ewiseblatt:
@lwander 
This executes for me, though the test itself is failing in my configuration (failure is within spinnaker). I think the failure is unrelated though still investigating.


lwander:
LGTM

PR #203, opened 2016-08-26
disabled google http load balancer tests

danielpeach:
@ewiseblatt please review


ewiseblatt:
For the record, this is because these endpoints havent been surfaced in deck yet so are not in use, and are undergoing a breaking change.

PR #207, opened 2016-09-02
Added pipeline test and test history.

ewiseblatt:
@lwander

### DIFF
--- a/spinnaker/spinnaker_system/google_front50_test.py
+++ b/spinnaker/spinnaker_system/google_front50_test.py
@@ -18,9 +18,11 @@
 import json
 import logging
 import sys
+import urllib
 
 from citest.base import ExecutionContext
 
+import citest.base
 import citest.gcp_testing as gcp
 import citest.json_contract as jc
 import citest.service_testing as st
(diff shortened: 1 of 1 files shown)


spinnaker/spinnaker_system/google_front50_test.py
```diff
@@ -158,32 +179,46 @@ def update_app(self):
...
+
+    for clause in gcs_builder.build().clauses:
+      contract.add_clause(clause)
+
+    f50_builder = st.http_observer.HttpContractBuilder(self.agent)
+    (f50_builder.new_clause_builder('History Records Changes')
+     .get_url_path('/default/applications/{app}/history'
+                   .format(app=self.TEST_APP))
+     .contains_path_value('[0]', self.app_history[0])
+     .contains_path_value('[1]', self.app_history[1]))
```
lwander: The flow with the `self.app_history` is a little confusing - what's going on here? I see you insert expected values above, but I'm not sure what you're searching for.

ewiseblatt: The historic entities were remembered properly.
Path '[N]' means the Nth element in a list. So this is checking the list reflects the value and ordering.

lwander: Oh I see, makes sense, thanks

lwander:
LGTM

PR #208, opened 2016-09-14
Added L7 LB upsert test.

jtk54:
@ewiseblatt please review. @duftler FYI. This is a pretty comprehensive test on L7 load balancer upsert without associated server groups. I'll add another set of tests in the future that test the L7 server group operations and validates L7 upsert when there are server groups attached.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ build/
 #*
 *.html
 *.log
+*.journal
+*.egg-info/
+venv/
--- a/spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
+++ b/spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
@@ -1,0 +1,681 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Test scenario for Gcp Http(s) Load Balancers.
+
+# Standard python modules.
+from OpenSSL import crypto
+import copy
+import json
+import sys
+import time
+import uuid
+
+# citest modules.
+from citest.base import ExecutionContext
+import citest
+import citest.gcp_testing as gcp
+import citest.json_predicate as jp
+import citest.service_testing as st
+
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+
+SCOPES = [gcp.COMPUTE_READ_WRITE_SCOPE]
+GCE_URL_PREFIX = 'https://www.googleapis.com/compute/v1/projects/'
+
+class GoogleHttpLoadBalancerTestScenario(sk.SpinnakerTestScenario):
+  '''Defines the tests for L7 Load Balancers.
+  '''
+
+  MINIMUM_PROJECT_QUOTA = {
+    'INSTANCE_TEMPLATES': 1,
+    'BACKEND_SERVICES': 3,
+    'URL_MAPS': 1,
+    'HEALTH_CHECKS': 1,
+    'IN_USE_ADDRESSES': 2,
+    'SSL_CERTIFICATES': 2,
+    'TARGET_HTTP_PROXIES': 1,
+    'TARGET_HTTPS_PROXIES': 1,
+    'FORWARDING_RULES': 2
+  }
+
+  MINIMUM_REGION_QUOTA = {
+      'CPUS': 2,
+      'IN_USE_ADDRESSES': 2,
+      'INSTANCE_GROUP_MANAGERS': 1,
+      'INSTANCES': 2,
+  }
+
+  @classmethod
+  def new_agent(cls, bindings):
+    '''Implements citest.service_testing.AgentTestScenario.new_agent.'''
+    agent = gate.new_agent(bindings)
+    agent.default_max_wait_secs = 600
+    return agent
+
+
+  def __init__(self, bindings, agent=None):
+    '''Constructor.
+
+    Args:
+      bindings: [dict] The data bindings to use to configure the scenario.
+      agent: [GateAgent] The agent for invoking the test operations on Gate.
+    '''
+    super(GoogleHttpLoadBalancerTestScenario, self).__init__(bindings, agent)
+
+    bindings = self.bindings
+
+    self.__lb_detail = 'lb'
+    self.TEST_APP = bindings['TEST_APP']
+    self.__lb_name = '{app}-{stack}-{detail}'.format(
+        app=bindings['TEST_APP'], stack=bindings['TEST_STACK'],
+        detail=self.__lb_detail)
+    self.__proto_hc = {
+      'name': 'basic-http-check',
+      'requestPath': '/',
+      'port': 80,
+      'checkIntervalSec': 2,
+      'timeoutSec': 1,
+      'healthyThreshold': 3,
+      'unhealthyThreshold': 4
+    }
+    self.__proto_upsert = {
+      'cloudProvider': 'gce',
+      'provider': 'gce',
+      'stack': bindings['TEST_STACK'],
+      'credentials': bindings['GCE_CREDENTIALS'],
+      'region': bindings['TEST_GCE_REGION'],
+      'loadBalancerType': 'HTTP',
+      'loadBalancerName': self.__lb_name,
+      'portRange': '80',
+      'defaultService': {
+        'name': 'default-backend-service',
+        'backends': [],
+        'healthCheck': self.__proto_hc,
+      },
+      'certificate': '',
+      'hostRules': [
+        {
+          'hostPatterns': ['host1.com', 'host2.com'],
+          'pathMatcher': {
+            'pathRules': [
+              {
+                'paths': ['/path', '/path2/more'],
+                'backendService': {
+                  'name': 'backend-service',
+                  'backends': [],
+                  'healthCheck': self.__proto_hc,
+                }
+              }
+            ],
+            'defaultService': {
+              'name': 'pm-backend-service',
+              'backends': [],
+              'healthCheck': self.__proto_hc,
+            }
+          }
+        }
+      ],
+      'type': 'upsertLoadBalancer',
+      'availabilityZones': {bindings['TEST_GCE_REGION']: []},
+      'user': '[anonymous]'
+    }
+
+
+  def _get_bs_link(self, bs):
+    '''Make a fully-formatted backend service link.
+    '''
+    return (GCE_URL_PREFIX + self.bindings['GCE_PROJECT'] +
+            '/global/backendServices/' + bs)
+
+
+  def _get_hc_link(self, hc):
+    '''Make a fully-formatted health check link.
+    '''
+    return (GCE_URL_PREFIX + self.bindings['GCE_PROJECT'] +
+            '/global/httpHealthChecks/' + hc)
+
+
+  def _set_all_hcs(self, upsert, hc):
+    '''Set all health checks in upsert to hc.
+    '''
+    upsert['defaultService']['healthCheck'] = hc
+    for host_rule in upsert['hostRules']:
+      path_matcher = host_rule['pathMatcher']
+      path_matcher['defaultService']['healthCheck'] = hc
+      for path_rule in path_matcher['pathRules']:
+        path_rule['backendService']['healthCheck'] = hc
+
+
+  def _add_contract_clauses(self, contract_builder, upsert):
+    '''Add the proper predicates to the contract builder for a given
+    upsert description.
+    '''
+    host_rules = upsert['hostRules'] # Host rules will be distinct.
+    backend_services = [upsert['defaultService']]
+    for host_rule in host_rules:
+      path_matcher = host_rule['pathMatcher']
+      backend_services.append(path_matcher['defaultService'])
+      for path_rule in path_matcher['pathRules']:
+        backend_services.append(path_rule['backendService'])
+    health_checks = [service['healthCheck'] for service in backend_services]
+
+    hc_clause_builder = (contract_builder
+                         .new_clause_builder('Health Checks Created', retryable_for_secs=30)
+                         .list_resource('httpHealthChecks'))
+    for hc in health_checks:
+      hc_clause_builder.contains_pred_list(
+        [
+          jp.PathEqPredicate('name', hc['name']),
+          jp.PathEqPredicate('requestPath', hc['requestPath']),
+          jp.PathEqPredicate('port', hc['port'])
+        ]
+      )
+
+    bs_clause_builder = (contract_builder.
+                         new_clause_builder('Backend Services Created', retryable_for_secs=30).
+                         list_resource('backendServices'))
+    for bs in backend_services:
+      bs_clause_builder.contains_pred_list(
+        [
+         jp.PathEqPredicate('name', bs['name']),
+         jp.PathEqPredicate('portName', 'http'),
+         jp.PathContainsPredicate('healthChecks[0]', self._get_hc_link(bs['healthCheck']['name']))
+        ]
+      )
+
+    url_map_clause_builder = (contract_builder
+                              .new_clause_builder('Url Map Created', retryable_for_secs=30)
+                              .list_resource('urlMaps'))
+    for hr in host_rules:
+      pred_list = []
+      pm = hr['pathMatcher']
+      pred_list.append(jp.AND([
+        jp.PathEqPredicate('name', self.__lb_name),
+        jp.PathEqPredicate('defaultService',
+                           self._get_bs_link(upsert['defaultService']['name'])),
+        jp.PathContainsPredicate('pathMatchers/defaultService',
+                                 self._get_bs_link(pm['defaultService']['name'])),
+      ]))
+      pred_list.append(
+        jp.AND([jp.PathContainsPredicate('hostRules/hosts',
+                                         host) for host in hr['hostPatterns']])
+      )
+      for pr in pm['pathRules']:
+        pred_list.append(
+         jp.PathContainsPredicate('pathMatchers/pathRules/service',
+                                  self._get_bs_link(pr['backendService']['name'])),
+        )
+        for path in pr['paths']:
+          pred_list.append(
+            jp.PathContainsPredicate('pathMatchers/pathRules/paths', path),
+          )
+      url_map_clause_builder.contains_pred_list(pred_list)
+
+    port_string = '443-443'
+    if upsert['certificate'] == '':
+      port_string = '%s-%s' % (upsert['portRange'], upsert['portRange'])
+
+    (contract_builder.new_clause_builder('Forwarding Rule Created', retryable_for_secs=30)
+     .list_resource('globalForwardingRules')
+     .contains_pred_list(
+       [
+         jp.PathEqPredicate('name', self.__lb_name),
+         jp.PathEqPredicate('portRange', port_string)
+       ]))
+
+    proxy_clause_builder = contract_builder.new_clause_builder('Target Proxy Created',
+                                                               retryable_for_secs=30)
+    self._add_proxy_clause(upsert['certificate'], proxy_clause_builder)
+
+
+  def _add_proxy_clause(self, certificate, proxy_clause_builder):
+    target_proxy_name = '%s-target-%s-proxy'
+    if certificate:
+      target_proxy_name = target_proxy_name % (self.__lb_name, 'https')
+      (proxy_clause_builder.list_resource('targetHttpsProxies')
+       .contains_pred_list([jp.PathEqPredicate('name', target_proxy_name)]))
+    else:
+      target_proxy_name = target_proxy_name % (self.__lb_name, 'http')
+      (proxy_clause_builder.list_resource('targetHttpProxies')
+       .contains_pred_list([jp.PathEqPredicate('name', target_proxy_name)]))
+
+
+  def upsert_full_load_balancer(self):
+    '''Upserts L7 LB with full hostRules, pathMatchers, etc.
+
+    Calls the upsertLoadBalancer operation with a payload, then verifies that
+    the expected resources are visible on GCP.
+    '''
+    hc = copy.deepcopy(self.__proto_hc)
+    hc['requestPath'] = '/'
+    hc['port'] = 80
+    upsert = copy.deepcopy(self.__proto_upsert)
+    self._set_all_hcs(upsert, hc)
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='upsert full http lb',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def upsert_min_load_balancer(self):
+    '''Upserts a L7 LB with the minimum description.
+    '''
+    upsert = copy.deepcopy(self.__proto_upsert)
+    upsert['hostRules'] = []
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='upsert min http lb', data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def delete_load_balancer(self):
+    '''Deletes the L7 LB.
+    '''
+    bindings = self.bindings
+    delete = {
+      'type': 'deleteLoadBalancer',
+      'cloudProvider': 'gce',
+      'loadBalancerType': 'HTTP',
+      'loadBalancerName': self.__lb_name,
+      'region': bindings['TEST_GCE_REGION'],
+      'regions': [bindings['TEST_GCE_REGION']],
+      'credentials': bindings['GCE_CREDENTIALS'],
+      'user': '[anonymous]'
+    }
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[delete],
+      description='Delete L7 Load Balancer: {0} in {1}:{2}'.format(
+        self.__lb_name,
+        bindings['GCE_CREDENTIALS'],
+        bindings['TEST_GCE_REGION'],
+      ),
+      application=self.TEST_APP
+    )
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    (contract_builder.new_clause_builder('Health Check Removed', retryable_for_secs=30)
+     .list_resource('httpHealthChecks')
+     .excludes_path_value('name', self.__proto_hc['name'])
+    )
+    (contract_builder.new_clause_builder('Url Map Removed',
+                                         retryable_for_secs=30)
+     .list_resource('urlMaps')
+     .excludes_path_value('name', self.__lb_name)
+    )
+    (contract_builder.new_clause_builder('Forwarding Rule Removed',
+                                         retryable_for_secs=30)
+     .list_resource('globalForwardingRules')
+     .excludes_path_value('name', self.__lb_name)
+    )
+
+    return st.OperationContract(
+      self.new_post_operation(
+        title='delete_load_balancer', data=payload, path='tasks'),
+      contract=contract_builder.build())
+
+
+  def change_health_check(self):
+    '''Changes the health check associated with the LB.
+    '''
+    upsert = copy.deepcopy(self.__proto_upsert)
+    hc = copy.deepcopy(self.__proto_hc)
+    hc['requestPath'] = '/changedPath'
+    hc['port'] = 8080
+    self._set_all_hcs(upsert, hc)
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='change health checks',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def change_backend_service(self):
+    '''Changes the default backend service associated with the LB.
+    '''
+    hc = copy.deepcopy(self.__proto_hc)
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    hc['name'] = 'updated-hc'
+    hc['requestPath'] = '/changedPath1'
+    hc['port'] = 8080
+
+    bs_upsert['defaultService']['healthCheck'] = hc
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='change backend services',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def add_host_rule(self):
+    '''Adds a host rule to the url map.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    hr = copy.deepcopy(bs_upsert['hostRules'][0])
+    hr['hostPatterns'] = ['added.host1.com', 'added.host2.com']
+    hr['pathMatcher']['pathRules'][0]['paths'] = ['/added/path']
+    bs_upsert['hostRules'].append(hr)
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='add host rule',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def update_host_rule(self):
+    '''Updates a host rule to the url map.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    hr = copy.deepcopy(bs_upsert['hostRules'][0])
+    hr['hostPatterns'] = ['updated.host1.com']
+    hr['pathMatcher']['pathRules'][0]['paths'] = ['/updated/path']
+    bs_upsert['hostRules'].append(hr)
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='update host rule',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def update_port_range(self):
+    '''Updates the port range on the forwarding rule.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    bs_upsert['portRange'] = '8080'
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='update host rule',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def add_cert(self, certname):
+    '''Add cert to targetHttpProxy to make it a targetHttpsProxy.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    bs_upsert['certificate'] = certname
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='add cert',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def add_security_group(self):
+    '''Associates a security group with the L7 load balancer.
+    '''
+    bindings = self.bindings
+    sec_group_payload = self.agent.make_json_payload_from_kwargs(
+      job=[
+        {
+          'allowed': [
+            {
+              'ipProtocol': 'tcp',
+              'portRanges': ['80-80']
+            },
+            {
+              'ipProtocol': 'tcp',
+              'portRanges': ['8080-8080']
+            },
+            {
+              'ipProtocol': 'tcp',
+              'portRanges': ['443-443']
+            }
+          ],
+          'backingData': {'networks': ['default']},
+          'cloudProvider': 'gce',
+          'application': self.TEST_APP,
+          'credentials': bindings['GCE_CREDENTIALS'],
+          'description': '',
+          'detail': 'http',
+          'ipIngress': [
+            {
+              'type': 'tcp',
+              'startPort': 80,
+              'endPort': 80,
+            },
+            {
+              'type': 'tcp',
+              'startPort': 8080,
+              'endPort': 8080,
+            },
+            {
+              'type': 'tcp',
+              'startPort': 443,
+              'endPort': 443,
+            }
+          ],
+          'name': self.__lb_name + '-rule',
+          'network': 'default',
+          'region': 'global',
+          'securityGroupName': self.__lb_name + '-rule',
+          'sourceRanges': ['0.0.0.0/0'],
+          'type': 'upsertSecurityGroup',
+          'user': '[anonymous]'
+        }
+      ],
+      description='Create a Security Group for L7 operations.',
+      application=self.TEST_APP
+    )
+    builder = gcp.GcpContractBuilder(self.gcp_observer)
+    (builder.new_clause_builder('Security Group Created',
+                                retryable_for_secs=30)
+     .list_resource('firewalls')
+     .contains_pred_list(
+       [
+         jp.PathContainsPredicate('name', self.__lb_name + '-rule') # TODO(jacobkiefer): Change name.
+       ]
+     )
+    )
+
+    return st.OperationContract(
+      self.new_post_operation(title='create security group',
+                              data=sec_group_payload, path='tasks'),
+      contract=builder.build()
+    )
+
+  def add_server_group(self):
+    '''Adds a server group to the L7 LB.
+    '''
+    time.sleep(60) # Wait for the L7 LB to be ready.
+    bindings = self.bindings
+    group_name = '{app}-{stack}-v000'.format(app=self.TEST_APP,
+                                             stack=bindings['TEST_STACK'])
+    policy = {
+      'balancingMode': 'UTILIZATION',
+      'listeningPort': 80,
+      'maxUtilization': 0.8,
+      'capacityScaler': 0.8
+    }
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[{
+        'cloudProvider': 'gce',
+        'application': self.TEST_APP,
+        'credentials': bindings['GCE_CREDENTIALS'],
+        'strategy':'',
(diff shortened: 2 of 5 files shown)


spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
+#
```
ewiseblatt: 2016

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+import citest.service_testing as st
+
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+
+HTTP_PROXY_PREFIX = 'target-http-proxy'
+HTTPS_PROXY_PREFIX = 'target-https-proxy'
+SCOPES = ['https://www.googleapis.com/auth/compute']
+GCE_URL_PREFIX = 'https://www.googleapis.com/compute/v1/projects/'
```
ewiseblatt: gcp.COMPUTE_READ_WRITE_SCOPE

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+      'INSTANCES': 2,
+  }
+
+  @classmethod
+  def new_agent(cls, bindings):
+    '''Implements citest.service_testing.AgentTestScenario.new_agent.'''
+    return gate.new_agent(bindings)
+
+
+  @classmethod
```
ewiseblatt: you dont need this method if you have no arguments.

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+              'defaultService': {
+                'name': 'pm-backend-service',
+                'backends': [],
+                'healthCheck': self.__default_hc,
+              }
+            }
+          }
+        ],
+        'type': 'upsertLoadBalancer',
+        'availabilityZones': {bindings['TEST_GCE_REGION']: []},
```
ewiseblatt: what does an empty list here mean?

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+    self.__default_hc = {
+      'name': 'basic-http-check',
+      'requestPath': '/',
+      'port': 80,
+      'checkIntervalSec': 1,
+      'timeoutSec': 1,
+      'healthyThreshold': 1,
+      'unhealthyThreshold': 1
+    }
+    self.__default_upsert = [
```
ewiseblatt: why is this a list? I'd make it just an object then wrap it in a list when you use it as the payload. Then you can remove all the [0] list indexing off referencing it and treat it as the object request it is.

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+              }
+            }
+          }
+        ],
+        'type': 'upsertLoadBalancer',
+        'availabilityZones': {bindings['TEST_GCE_REGION']: []},
+        'user': '[anonymous]'
+      }
+    ]
+    self.__default_delete = [
```
ewiseblatt: make this just a single instance and wrap it in a list where the payload wants a list.

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+      }
+    ]
+
+
+  def _make_bs_link(self, bs):
+    '''Make a fully-formatted backend service link.
+    '''
+    return (GCE_URL_PREFIX + self.bindings['GCE_PROJECT'] +
+            '/global/backendServices/' + bs)
+
```
ewiseblatt: I think these two methods (along with GCE_URL_PREFIX) belong in gcp_testing, probably gcp_compute_agent since these pertain to the Compute service and that agent is where knowledge of Compute lives.

I'm not sure offhand what the API would look like but suspect you would do a discovery document lookup as I do for the methods and resolve the variables into the URL. So assuming the discovery document tells you how to get the resource link (which I think is just the GET URL), this would look like:
agent.getResourceUrl(project=self.bindings['GCE_PROJECT'], resource='backendServices', resource_id=bs)

The knowledge of the actual URL comes from the discovery document, which comes directly from GCP at runtime and the variables are bound in by the caller.

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+    full_hc_name = self.__default_hc['name']
+    hc_link = self._make_hc_link(self.__default_hc['name'])
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    (contract_builder.new_clause_builder('Health Checks Created',
+                                         retryable_for_secs=30)
+     .list_resource('httpHealthChecks')
+     .contains_pred_list(
+       [
```
ewiseblatt: Here and below, I think often you want PathEqPredicate to look for exact values.
The full_hc_name is contains since it is a subset.
The port is an integer so Contains is the same as Eq.
PathContainsPredicate('requestPath', '/') just means it has a slash in it, not it is the root. Is that what you meant?

A lot of the clauses below are using what looks like exact string literals, so Eq is more precise to what you mean for those.

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+         jp.PathContainsPredicate('port',
+                                  int(self.__default_upsert[0]['portRange'])),
+         jp.PathContainsPredicate('healthChecks[0]', hc_link),
+       ]
+     )
+    )
+    (contract_builder.new_clause_builder('Url Map Created',
+                                        retryable_for_secs=30)
+     .list_resource('urlMaps')
+     .contains_pred_list(
```
ewiseblatt: I wonder if this would be easier to read/write and report if you could say
 expect = {
    'name': jp.CONTAINS(self.__lb.name),
    'defaultService' : jp.EQUALS(self._make_bs_link('default-backend-service'),
     ...
}
.contains_pred(jp.PathMatches('', expect))


The way you are doing it has the most precise reporting because you can see which constraints passed and which failed. Passing in a dictionary currently reports all or none but also forces all the elements to have the same relation. The above allows you to specify an object profile with independent field comparators. We should be able to get the same reporting granularity and expression while also knowing that all the predicates are related by refering to the same object whereas a pred_list could refer to multiple objects or conditions and not just PathPredicates within an object.

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+    return st.OperationContract(
+      self.new_post_operation(title='upsert full http lb',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+  def upsert_min_load_balancer(self):
+    '''Upserts a L7 LB with the minimum description.
+    '''
+    min_upsert = copy.deepcopy(self.__default_upsert)
```
ewiseblatt: I think if default_upsert were just an object rather than a list then min_upsert would be dict(self.__default_upsert)

spinnaker/spinnaker_system/google_http_lb_test.py
```diff
@@ -0,0 +1,1256 @@
...
+    cert.sign(key, 'sha1')
+
+    compute_agent = gcp.GcpComputeAgent.make_agent(scopes=SCOPES,
+                                                   credentials_path=bindings['GCE_CREDENTIALS_PATH'])
+    ssl_cert = {
+      'name': FIRST_CERT,
+      'privateKey': crypto.dump_privatekey(crypto.FILETYPE_PEM, key),
+      'certificate': crypto.dump_certificate(crypto.FILETYPE_PEM, cert)
+    }
+    compute_agent.invoke_resource(ExecutionContext(), # No predicates against this agent, context is empty.
```
ewiseblatt: Use the same execution context for all of the setup.
It doenst matter right now, but if the context were to be used, you'd probably want the same one because the context (setUpClass) is the same.

spinnaker/spinnaker_system/google_http_lb_test.py
ewiseblatt: just to be clear, you cannot say that now. I'm speculating on whether I should enable you to.

jtk54:
@ewiseblatt I addressed most of your comments. I'll refactor the link building into the `GCPAgent` in a later PR.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+      title,
+      scenario.gcp_observer,
+      project_quota=upsert.GoogleHttpLoadBalancerTestScenario.MINIMUM_PROJECT_QUOTA,
+      regions=[(managed_region,
+                upsert.GoogleHttpLoadBalancerTestScenario.MINIMUM_REGION_QUOTA)]
+    )
+    if not verify_results:
+      raise RuntimeError('Insufficient Quota: {0}'.format(verify_results))
+
+    key = crypto.PKey()
```
ewiseblatt: move this block plus the ssl_cert declaration into a "make_ssl_cert(name)" function.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+      'name': cls.FIRST_CERT,
+      'privateKey': crypto.dump_privatekey(crypto.FILETYPE_PEM, key),
+      'certificate': crypto.dump_certificate(crypto.FILETYPE_PEM, cert)
+    }
+    compute_agent.invoke_resource(context, # No predicates against this agent, context is empty.
+                                  'insert',
+                                  resource_type='sslCertificates',
+                                  project=bindings['GCE_PROJECT'],
+                                  body=ssl_cert)
+    ssl_cert['name'] = cls.SECOND_CERT
```
ewiseblatt: This is confusing. Pass body=make_ssl_cert(cls.FIRST_CERT) above and body=make_ssl_cert(cls.SECOND_CERT) below.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+class GoogleHttpLoadBalancerTest(st.AgentTestCase):
+  '''Test fixture for Http LB test.
+  '''
+
+  FIRST_CERT = ''
+  SECOND_CERT = ''
+
+  @classmethod
+  def setUpClass(cls):
+    cls.FIRST_CERT = 'first-cert' + str(uuid.uuid4())
```
ewiseblatt: Given that it appears you are adding these into the GCE project, you should try to name them with the TEST_APP so the side effects can be traced back to this test. Especially if things go wrong and they are leaked.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+                                  project=bindings['GCE_PROJECT'],
+                                  body=ssl_cert)
+
+
+  @classmethod
+  def tearDownClass(cls):
+    runner = citest.base.TestRunner.global_runner()
+    bindings = runner.bindings
+    context = ExecutionContext()
+    compute_agent = gcp.GcpComputeAgent.make_agent(scopes=SCOPES,
```
ewiseblatt: I've been wanting to generalize the concept for cleaning up after scenarios but havent gotten to putting it in.

In the mean time, what do you think about having a class variable
cleanup_lambdas = []

Then in setupUpClass
cleanup_lamdas.append(lambda : compute_agent.invoke_resource(context, 'delete', ...))

Then in teardownClass
for func in cleanup_lambdas:
  func()

I think that would provide better encapsulation so knowledge about what needs to be cleaned up (and how to do it) is localized to where it is created.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+  @property
+  def scenario(self):
+    return citest.base.TestRunner.global_runner().get_shared_data(
+        upsert.GoogleHttpLoadBalancerTestScenario)
+
+  def test_a_upsert_lb(self):
+    self.run_test_case(self.scenario.upsert_full_load_balancer())
+
```
ewiseblatt: I thought we'd agreed that test_a and test_b were redundant to test_d and test_o

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+  def test_k_update_port_range(self):
+    self.run_test_case(self.scenario.update_port_range())
+
+  def test_l_add_cert(self):
+    self.run_test_case(self.scenario.add_cert(self.__class__.FIRST_CERT))
+
+  def test_m_change_cert(self):
+    self.run_test_case(self.scenario.change_cert(self.__class__.SECOND_CERT))
+
+  def test_n_delete_server_group(self):
```
ewiseblatt: FWIW, I'd been naming delete operations from the back to the front so that future inserts/removes would be easier.

Really I'd like to be able to specify that the test suite built from this leaves the order in declaration order rather than doing a lexical sort. I was never able to get that working. If you know how to do/fix that, then that would be great.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+    self.run_test_case(self.scenario.delete_server_group())
+
+  def test_o_delete_lb(self):
+    self.run_test_case(self.scenario.delete_load_balancer())
+
```
ewiseblatt: Wait, you never defined the scenario and tests. I see them in the other file below, but that has tests too. Is this file obsolete? If not, please explain.

spinnaker/spinnaker_system/google_http_upsert_test.py
```diff
@@ -0,0 +1,891 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
ewiseblatt: Could you rename this module so it more clearly relates to load balancing (even lb)?

spinnaker/spinnaker_system/google_http_upsert_test.py
```diff
@@ -0,0 +1,891 @@
...
+import citest.service_testing as st
+
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+
+HTTP_PROXY_PREFIX = 'target-http-proxy'
+HTTPS_PROXY_PREFIX = 'target-https-proxy'
+SCOPES = [gcp.COMPUTE_READ_WRITE_SCOPE]
+GCE_URL_PREFIX = 'https://www.googleapis.com/compute/v1/projects/'
```
ewiseblatt: Couldnt you add a gcp_agent.getResourceUrl() method and use the URL from the get method from the discovery document in the gcp_compute_agent as I'd suggested offline?

spinnaker/spinnaker_system/google_http_upsert_test.py
```diff
@@ -0,0 +1,891 @@
...
+  @classmethod
+  def new_agent(cls, bindings):
+    '''Implements citest.service_testing.AgentTestScenario.new_agent.'''
+    agent = gate.new_agent(bindings)
+    agent.default_max_wait_secs = 600
+    return agent
+
+
+  @classmethod
+  def initArgumentParser(cls, parser, defaults=None):
```
ewiseblatt: It looks like you added an update after you pinged me and I started looking at this.
I'm going to send this so far and look at the next update.
Disregard my comments if this is in fact old.

.gitignore
```diff
@@ -11,3 +11,6 @@ build/
 #*
 *.html
 *.log
+*.journal
+citest.egg-info/
+venv/
```
jtk54: citest.egg-info is a resulting file from `pip install -r requirements.txt` - it's where the python package info is kept. These are for ease of development, before I couldn't rebase properly without clearing my local `venv` state.

spinnaker/spinnaker_system/google_http_upsert_test.py
jtk54: Shooting for that in a future PR.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
jtk54: Seems like a bit of overhead that accomplishes the same thing as having the `tearDownClass`. What is the difference between queuing functions to run and having an explicit teardown block and running the same functions you would queue?

The `context`'s utility would be the determining factor here, otherwise it's very similar to what's available.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
jtk54: They both use the same scenario, one is a test with a populated load balancer, one is empty.

.gitignore
ewiseblatt: You dont need to .git-ignore it. Just dont add it to begin with.
If this is a problem, perhaps add *.egg-info since none of it would go into git anyway.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
ewiseblatt: The difference is encapsulation and locality. You can associate the deletion with the creation  and share the agent/credentials/scopes/key names.

It might not be a good idea. Just suggesting it might.

Also having that facility available could make it easy to ensure you clean up supporting artifacts after other tests if you dont need a delete test where it would be appropriate to do.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
ewiseblatt: Ah. This wasnt clear. Maybe the fixtures should decorate something like "WithServices" and "WithoutServices", and commented (assuming I didnt miss where it was).

jtk54:
@ewiseblatt addressed comments and renamed test files.

spinnaker/spinnaker_system/google_http_upsert_server_test.py
jtk54: I added to the module comments and explained a bit more, also added a comment in the `scenario` property function.

spinnaker/spinnaker_system/google_http_lb_upsert_server_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+SCOPES = [gcp.COMPUTE_READ_WRITE_SCOPE]
+
+class GoogleHttpLoadBalancerTest(st.AgentTestCase):
+  '''Test fixture for Http LB test.
+  '''
+
+  FIRST_CERT = ''
+  SECOND_CERT = ''
```
ewiseblatt: are these certs only used by one test? What if you wrapped each of those tests so it creates the certs, runs the test, then deletes the cert (in a finally clause). Or if the cert needs to be valid over the future lifetime of the thing referencing it, then having the test add a deletion lambda after it creates the cert woudl also eliminate needing these class variables and access points spread across several different methods.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+
+HTTP_PROXY_PREFIX = 'target-http-proxy'
+HTTPS_PROXY_PREFIX = 'target-https-proxy'
+SCOPES = [gcp.COMPUTE_READ_WRITE_SCOPE]
+GCE_URL_PREFIX = 'https://www.googleapis.com/compute/v1/projects/'
+
+class GoogleHttpLoadBalancerTestScenario(sk.SpinnakerTestScenario):
```
ewiseblatt: Since this is shared across different fixtures, maybe put it in its own file?

spinnaker/spinnaker_system/google_http_lb_upsert_server_test.py
```diff
@@ -0,0 +1,199 @@
...
+# group attached.
+
+# Standard python modules.
+from OpenSSL import crypto
+import copy
+import json
+import sys
+import time
+import uuid
+import google_http_upsert_test as upsert
```
ewiseblatt: from google_http_upsert_test import GoogleHttpLoadBalancerTestScenario

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+import citest
+import citest.gcp_testing as gcp
+import citest.json_predicate as jp
+import citest.service_testing as st
+
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+
+HTTP_PROXY_PREFIX = 'target-http-proxy'
```
ewiseblatt: These seem to only be used in one place. Why have a symbol for it?

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    agent = gate.new_agent(bindings)
+    agent.default_max_wait_secs = 600
+    return agent
+
+
+  @classmethod
+  def initArgumentParser(cls, parser, defaults=None):
+    '''Initialize command line argument parser.
+
+    Args:
```
ewiseblatt: this method isnt needed.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    self.TEST_APP = bindings['TEST_APP']
+    self.__lb_name = '{app}-{stack}-{detail}'.format(
+        app=bindings['TEST_APP'], stack=bindings['TEST_STACK'],
+        detail=self.__lb_detail)
+    self.__proto_hc = {
+      'name': 'basic-http-check',
+      'requestPath': '/',
+      'port': 80,
+      'checkIntervalSec': 1,
+      'timeoutSec': 1,
```
ewiseblatt: I like to make these values different so if the code transposes them you'll catch it.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    host_rules = upsert['hostRules'] # Host rules will be distinct.
+    backend_services = [upsert['defaultService']]
+    for host_rule in host_rules:
+      path_matcher = host_rule['pathMatcher']
+      backend_services.append(path_matcher['defaultService'])
+      for path_rule in path_matcher['pathRules']:
+        backend_services.append(path_rule['backendService'])
+    health_checks = [service['healthCheck'] for service in backend_services]
+
+    hc_clause_builder = contract_builder.new_clause_builder('Health Checks Created',
```
ewiseblatt: I'd combine these two to
hc_clause_bulider = contract_builder.new()
   .list_resource('httpHealthChecks')

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+        [
+          jp.PathEqPredicate('name', hc['name']),
+          jp.PathEqPredicate('requestPath', hc['requestPath']),
+          jp.PathEqPredicate('port', hc['port'])
+        ]
+      )
+
+    bs_clause_builder = contract_builder.new_clause_builder('Backend Services Created',
+                                                            retryable_for_secs=30)
+    bs_clause_builder = bs_clause_builder.list_resource('backendServices')
```
ewiseblatt: I'd combine these.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+        [
+         jp.PathEqPredicate('name', bs['name']),
+         jp.PathEqPredicate('portName', 'http'),
+         jp.PathContainsPredicate('healthChecks[0]', bs['healthCheck']['name'])
+        ]
+      )
+
+    url_map_clause_builder = contract_builder.new_clause_builder('Url Map Created',
+                                                                 retryable_for_secs=30)
+    url_map_clause_builder = url_map_clause_builder.list_resource('urlMaps')
```
ewiseblatt: I'd move these two to #245 and combine them all together.
I might also wrap the line before the '=' to avoid some line wraps in the clauses.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    backend_services = [upsert['defaultService']]
+    for host_rule in host_rules:
+      path_matcher = host_rule['pathMatcher']
+      backend_services.append(path_matcher['defaultService'])
+      for path_rule in path_matcher['pathRules']:
+        backend_services.append(path_rule['backendService'])
+    health_checks = [service['healthCheck'] for service in backend_services]
+
+    hc_clause_builder = contract_builder.new_clause_builder('Health Checks Created',
+                                                            retryable_for_secs=30)
```
ewiseblatt: I prefer to wrap the line after the "(" so that both arguments are on the same line and it is indented closer to the left margin. It makes the flow easier to read plus keeps the block from looking fragmented with the vertical whitespace suggesting a different block. Especially in git's web UI.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+        pred_list.append(
+         jp.PathContainsPredicate('pathMatchers/pathRules/service',
+                                  self._make_bs_link(pr['backendService']['name'])),
+        )
+        for path in pr['paths']:
+          pred_list.append(
+            jp.PathContainsPredicate('pathMatchers/pathRules/paths', path),
+          )
+      url_map_clause_builder.contains_pred_list(
+        pred_list
```
ewiseblatt: The wrap here isnt necessary, especially if you combine the clause_builder statements together into one statement.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+      )
+
+    fr_clause_builder = contract_builder.new_clause_builder('Forwarding Rule Created',
+                                                            retryable_for_secs=30)
```
ewiseblatt: combine

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    fr_clause_builder = fr_clause_builder.list_resource('globalForwardingRules')
+
+    port_string = '443-443'
+    if upsert['certificate'] == '':
+      port_string = '%s-%s' % (upsert['portRange'], upsert['portRange'])
+
+    fr_clause_builder.contains_pred_list(
```
ewiseblatt: combine this with the above clauses. Move the variable definitions before #249

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+      [
+         jp.PathEqPredicate('name', self.__lb_name),
+         jp.PathEqPredicate('portRange', port_string)
+      ]
+    )
+
+    proxy_clause_builder = contract_builder.new_clause_builder('Target Proxy Created',
+                                                               retryable_for_secs=30)
+    if upsert['certificate'] == '':
```
ewiseblatt: if upsert['certificate']:  (and reverse cases)

ewiseblatt: actually, both clauses are the same except for a couple parameters. I'd either create a parameterized function or define the variables here.

if upsert['certificate']:
  makeTargetProxyCreatedClause('targetHttpProxies', HTTP_PROXY_PREFIX)
else:
  makeTargetProxyCreatedClause('targetHttpsProxies', HTTPS_PROXY_PREFIX)

but since *PROXY_PREFIX isnt actually referenced anywhere, this is just:
  kind = 'Https' if clause else 'Http'
  target_proxy_name = '%s-target-%s-proxy' % (self.__lb_name, kind.lower())
  contract_builder.new_clause_builder('Target Proxy Created', retryable_for_secs=30)
    .list_resource('target{0}Proxies'.format(kind))
    .contains_path_eq('name', target_proxy_name)

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[upsert],
+      descriptiopn='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    hc_link = self._make_hc_link(hc['name'])
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._make_contract_clauses(contract_builder, upsert)
```
ewiseblatt: I dont like make here for personal bias reasons of wanting you to return something rather than have a side effect. Could you instead call this _add_standard_contract_clauses(...)

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
ewiseblatt: Hmm, I see I didnt do that in any of the tests I wrote. I wonder why not. Here seems even worse because the clause construction is spread out so the block boundaries arent as clear, and I'm reading this unfamiliar code in git's UI which is < 80 columns so wraps weird.

I might change my wrapping as well now that I'm aware of it...
Usually I try to keep entire parameter lists together.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    upsert = copy.deepcopy(self.__proto_upsert)
+    upsert['hostRules'] = []
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[upsert],
+      descriptiopn='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    hc_link = self._make_hc_link(self.__proto_hc['name'])
```
ewiseblatt: Is this used?

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    hc['port'] = 80
+    self._set_all_hcs(upsert, hc)
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[upsert],
+      descriptiopn='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    hc_link = self._make_hc_link(hc['name'])
```
ewiseblatt: Is this used?

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+      'user': '[anonymous]'
+    }
+
+
+  def _make_bs_link(self, bs):
+    '''Make a fully-formatted backend service link.
+    '''
+    return (GCE_URL_PREFIX + self.bindings['GCE_PROJECT'] +
+            '/global/backendServices/' + bs)
+
```
ewiseblatt: I prefer "get_*_link". Your other "make" have side effects or are more factory like. This is conceptually a lookup or derivation.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+        title='delete_load_balancer', data=payload, path='tasks'),
+      contract=contract_builder.build())
+
+
+  def change_health_check(self):
+    '''Changes the health check associated with the LB.
+    '''
+    upsert = copy.deepcopy(self.__proto_upsert)
+    hc = copy.deepcopy(self.__proto_hc)
+    hc['requestPath'] = '/hello'
```
ewiseblatt: Is there something significant about hello? If not, how about '/changedPath'

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    )
+
+
+  def change_backend_service(self):
+    '''Changes the default backend service associated with the LB.
+    '''
+    hc = copy.deepcopy(self.__proto_hc)
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    hc['name'] = 'updated-hc'
+    hc['requestPath'] = '/hello'
```
ewiseblatt: Could this changed health check and the previous one be confused somehow (a bug where you updated the wrong thing here and updated what the previous test was checking for, or vice versa)? If so then you might want to make these different so that the failures arent masked by the other test.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+    hc['port'] = 8080
+
+    bs_upsert['defaultService']['healthCheck'] = hc
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      descriptiopn='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    hc_link = self._make_hc_link(hc['name'])
```
ewiseblatt: These 'makes' are really 'gets' and you arent using the values anywhere. What is the point of these?

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+      self.new_post_operation(title='change backend services',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
+
+  def add_host_rule(self):
+    '''Adds a host rule to the url map.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
```
ewiseblatt: This would flow better moving #442 after #445

As a rule of thumb, I like to try to keep locality of reference. You create an object, then prepare an attribute value, then add that new attribute to the object you'd created before. If you change the order, you prepare a value, then create an object and set the value on it. The difference is subtle, but magnified a bit more in that you then go on to use that object and forget about the value rather than coming back to it.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+      )
+
+
+  def upsert_full_load_balancer(self):
+    '''Upserts L7 LB with full hostRules, pathMatchers, etc.
+
+    Calls the upsertLoadBalancer operation with a payload, then verifies that
+    the expected resources are visible on GCP.
+    '''
+    upsert = copy.deepcopy(self.__proto_upsert)
```
ewiseblatt: This woudl flow better moving #290 after #293

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+      contract=contract_builder.build()
+    )
+
+
+  def update_host_rule(self):
+    '''Updates a host rule to the url map.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    hr = copy.deepcopy(bs_upsert['hostRules'][0])
+    hr['hostPatterns'] = ['golang.com']
```
ewiseblatt: I presume these patterns are arbitrary. Maybe name the hosts from the tests, like updated.host.com as opposed to original.host.com. This will also make it clearer when looking at logs and other state especially when a test fails and you want to know which value you have and where it came from.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+
+  def update_port_range(self):
+    '''Updates the port range on the forwarding rule.
+    '''
+    bs_upsert = copy.deepcopy(self.__proto_upsert)
+    bs_upsert['portRange'] = '8080'
+
+    payload = self.agent.make_json_payload_from_kwargs(
+      job=[bs_upsert],
+      descriptiopn='Upsert L7 Load Balancer: ' + self.__lb_name,
```
ewiseblatt: s/descriptiopn/description/g

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._make_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='add cert',
+                              data=payload, path='tasks'),
+      contract=contract_builder.build()
+    )
+
```
ewiseblatt: I cannot tell the difference between add and change methods other than the title given to the post operation.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -0,0 +1,894 @@
...
+     .excludes_path_value('name', group_name))
+
+    return st.OperationContract(
+      self.new_post_operation(
+        title='delete server group', data=payload, path='tasks'),
+      contract=builder.build()
+    )
+
+
+class GoogleHttpLoadBalancerTest(st.AgentTestCase):
```
ewiseblatt: I dont see a difference between this and the other GoogleHttpLoadBalancerTest defined above.
Could you have one class derive from the other (probably that from this since that uses the scenario defined here) and have the specialized class just clarify the difference. Hopefully it is very small.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
jtk54: #245 is run per host rule in the description. It doesn't make sense to move this statement there.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
jtk54: One creates and adds a server group at the beginning.

jtk54:
@ewiseblatt addressed comments, split out scenario into separate file.

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
ewiseblatt: I hate this github UI. Critique and Mondrian were so much better.
I'm not sure what I was seeing yesterday. I must not have seen the loop. You're right it doesnt make sense. In fact I think I missed that all these other loops are nested.

If I were doing a loop like this and couldnt write helper functions, i'd probably indent the block under #223 by 4 spaces rather than two (and add disable pylint indentation checking for that scope) to emphasize the scoping difference since it is so long.

But in this case what I'd do (but intentionally bit my tongue earlier) is create list of jp.AND predicate, build up that predicate with all the conditions, then add that predicate with build.add_constraint(host_result_predicate)

for hr in host_rules:
   pred_list = []
   pred_list.append(jp.AND(#226 - #228)
   pred_list.append(jp.AND( #244 for host in hr['hostPatterns']))
   ...
   builder.add_constraint(pred_list)

The above, I think, is clearer what you are saying. It adds another level of ANDs  ((a and b) AND (c and d)) rather than (a and b and c and d). I think it is easier to read and conceptualize, and might be easier to look at the reporting because you can see the context of the failure to the subexpression group rather than top-level element in a long chain.

Even without such a long nesting of four loops, when I thought it was just one, I'd still use the builder.add_constraint on a list. I think it makes the code cleaner since you can create the clause with the observation and predicates in a single literary statement.

jtk54:
@ewiseblatt Tried to fix up the for loops in `_add_contract_clauses` a bit for readability.

ewiseblatt:
I think I forgot to mention that you should consolidate the two fixtures.
If the only difference is adding the server group, you should be able to
subclass one from the other.
In the specialized class implement the class setup and teardown to call the
base class one (it adds nothing else)
then just add the one server group test (and delete test if applicable) and
I think you are good to go.
I think since it is sorting the tests lexigraphically that the derived
fixture tests will be injected into the right order.

ewiseblatt:
If you do that, dont squash yet so I can start reviewing now without
getting messed up.

ewiseblatt:
Also, I keep forgetting to mention that if you havent run pylint, do that
too.
I havent been as strict about doing that in the tests as in the core.
You can disable some of the warnings like documentation and maybe naming
and stuff.
I dont care about being strict with style there, rather want to remove the
noise to find bugs in the tests themselves.

PYTHONPATH=.:spinnaker pylint --rcfile=pylint.rc  <your files>

spinnaker/spinnaker_system/google_http_lb_upsert_test.py
```diff
@@ -134,31 +140,34 @@ def scenario(self):
...
     self.run_test_case(self.scenario.upsert_min_load_balancer())
 
-  def test_p_upsert_full_lb(self):
+  def test_d_upsert_full_lb(self):
     self.run_test_case(self.scenario.upsert_full_load_balancer())
 
-  def test_s_change_hc(self):
+  def test_e_add_security_group(self):
+    self.run_test_case(self.scenario.add_security_group())
+
```
ewiseblatt: Mention 'f' is reserved for the other fixture to inject its test into this part of the sequence.

The naming here is going to be brittle, especially the way we're interleaving classes and given the number of tests.

It looks like there are 3 phases of tests (which is typical)
creation
usage
deletion

And furthermore, the usage is  partitioned into modifying the different aspects.
I wonder if you added another grouping decorator then that would reduce the brittleness to firewall changes within a group.

For example, adding another thing in the creation phase wouldnt cascade through the rest of the tests. If it could be added at the end, it would have no impact at all.

Assuming ordering doesnt matter (or you could decorate again after the phase name)
def test_phaseA_creation_upsert_min_...
def test_phaseB_creation_upsert_full_...

The modifications are going to be order sensitive
def test_phaseC_modify_a_...

What do you think?

You can leave a whole in a phase, or a whole for a completely new phase, for use by the subclass depending on how you view it.

jtk54: I think having one more dimension here could be helpful, but we would have to provide pretty detailed examples on how they work. Could we use a decorator rather than encode the phases into the name? I actually think using a decorator for both would be useful -- something like `@phase('create')` or `@create_phase`, etc, then an ordering like `@order(1)`.

ewiseblatt:
I'm sorry, when I said decorator, I meant text in the name, like the string
I gave, not a language construct.
For one, I'm not that sophisticated with python yet.

- eric

jtk54:
My guess on the missing OpenSSL header file is that OS X has its own particular flavor of the library that doesn't quite work with `pip`. I'll test this out on an Ubuntu 14.04 vm similar to the test vm.

jtk54:
I was able to install on a fresh 14.04 VM without issues. Maybe we can write a test Jenkins job on our server to try it out.

ewiseblatt:
I'll patch it into my server.

This is still a concern because it makes osx unusable. I'm about to migrate
google/citest/spinnaker to spinnaker/spinnaker/testing where we cant assume
the audience is us and I know people work off osx.

I wont let that hold up submitting this PR initially, but am going to ask
you to look into resolving that so might as well now (if you can) while I
look at an actual execution (on ubuntu).

spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
```diff
@@ -0,0 +1,681 @@
...
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='update host rule',
```
ewiseblatt: update port range

spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
```diff
@@ -0,0 +1,681 @@
...
+      job=[bs_upsert],
+      description='Upsert L7 Load Balancer: ' + self.__lb_name,
+      application=self.TEST_APP
+    )
+
+    contract_builder = gcp.GcpContractBuilder(self.gcp_observer)
+    self._add_contract_clauses(contract_builder, bs_upsert)
+
+    return st.OperationContract(
+      self.new_post_operation(title='add cert',
```
ewiseblatt: You also use this function for a test that changes the cert so parameterize the title and pass it into add_cert.

spinnaker/spinnaker_system/google_http_lb_upsert_server_test.py
```diff
@@ -0,0 +1,195 @@
...
+                                  project=bindings['GCE_PROJECT'],
+                                  body=cls.make_ssl_cert(cls.SECOND_CERT, bindings))
+
+
+  @classmethod
+  def tearDownClass(cls):
+    runner = citest.base.TestRunner.global_runner()
+    bindings = runner.bindings
+    context = ExecutionContext()
+    compute_agent = gcp.GcpComputeAgent.make_agent(scopes=SCOPES,
```
ewiseblatt: If you look at the report, you can see these initializers in setup and tear down. It's ok, but clutters it up since it isnt interesting.

What I've done is add the constructor to _do_init_bindings(self) on the scenario and then reference it here. The _do_init_bindings is called within the "Configure Cloud Bindings" logging context, which puts them out of normal view.

jtk54: I'm not sure I can do that from the cleanup/teardown methods since they are class methods. There's no `self` to access the scenario. I couldn't find any workarounds, do you have any ideas?

ewiseblatt: Oh yea. I forgot about that. I still do it there. Since we put all the tests as methods in the same class, rather than as one method in different classes, the constructor and setupClass are essentially equivalent (except the latter is called before you instantiate, but that doesnt matter). In your case, you are subclassing the fixture so that could have implications with it getting called twice. But we're calling it twice anyway since each class is doing it. So I think given our pattern, this is equivalent. Overriding the setupTestCase is easier than overriding __init__ because __init__ takes some awkward parameters. That's why I added _do_init_bindings(). Also, _do_init_bindings() is called later, after some of the bindings are done.

Actually, that's even more reason to be deferring it.

ewiseblatt:
Fix your .gitconfig on your laptop and repush your change.

jtk54:
Fixed up my git credentials an re-synced.

jtk54:
The setup worked for me with `pip install -r requirements.txt`, which I assumed was equivalent to `easy_install .`. Turns out it isn't. Can you try to run this suite from your mac with the `pip` command?

ewiseblatt:
Yea, that worked. Turns out I had even documented that in the spinnaker/README.md but forgot that's how you do it.

jtk54:
Got it, good stuff.

ewiseblatt:
I asked you to squash, but I dont see that in this request in the 
history so am resending.

ewiseblatt:
Dang, did you just merge unsquashed?

ewiseblatt:
Any chance you can back those changes out before the repo is cloned?

jtk54:
No, I just squash-merged. It's one commit with the commit messages merged from this PR merged into one message.

ewiseblatt:
Ok. I'd just checked the commit history, but was in this branch, not hte master.
Thanks. I'm about to migrate the repo (with history) into spinnaker/ and dont want all those entries.

PR #210, opened 2016-09-22
Changed operation status_collector and cleanup functions.

ewiseblatt:
@jtk54 
The old status_collector (and cleanup) functions predated contexts.
Changed status_collector to status_extractor, whose responsibility can
be to move values from the status into the context. in the case of spinnaker,
this could traverse the complex json data to extract out a couple individual
fields that might be referenced by a test.

Cleanup just takes a context. The run_test_case method adds the verify results
into the context so they are still available (and the status was already there).

I also renamed the keys that the run_test_case writes into the context to be
camel case to match the type names. I dont think anything was referencing this
yet.

### DIFF
--- a/tests/service_testing/agent_test_case_test.py
+++ b/tests/service_testing/agent_test_case_test.py
@@ -22,7 +22,7 @@ import citest.service_testing as st
 import citest.json_contract as jc
 
 from citest.base import ExecutionContext
-from .fake_agent import (
+from fake_agent import (
     FakeAgent,
     FakeOperation,
     FakeStatus)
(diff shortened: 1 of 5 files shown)


tests/service_testing/agent_test_case_test.py
```diff
@@ -91,31 +91,79 @@ def test_assertVerifyResults_failed(self):
...
+      execution_context = ExecutionContext() if with_context else None
+
+      @staticmethod
+      def status_extractor(status, context):
+        if with_context:
+          # Verify this is the context we injected.
+          self.assertEquals(HelperClass.execution_context, context)
+        self.assertIsNotNone(context.get('OperationStatus', None))
+        self.assertIsNone(context.get('GOT_STATUS', None))
+        context['GOT_STATUS'] = status
```
jtk54: What else can we do with the status? How do we get to the json payload to unpack?

ewiseblatt: It's already there in the status. This is just the interface. Actual status is specialized by the agent that knows how to interpret it. The HttpAgent has a "detail" attribute which is the output of the server, though it . The SpinnakerAgent decodes this into json and provides an attribute "detail_doc". So for tests using SpinnakerAgent (e.g. using the GateAgent), the status of these operations will have a detail_doc with a dictionary form of the JSON response.

jtk54:
LGTM

tests/service_testing/agent_test_case_test.py
ewiseblatt: The bake_and_deploy_test uses this to identify the image to delete. I havent updated how it does that to take advantage of execution context to make that simpler. It pulls the status directly and walks through it. I'd like to add helper functions in the spinnaker agents that might be able to decode at least some of the status to populate the context with access points closer to the leaf that the test is looking.

PR #211, opened 2016-09-22
Fixed front50 retry params

ewiseblatt:
@jtk54


ewiseblatt:
I'm saying it doesnt matter (will be once a second). Before I had the wrong
API. The retries is for operations, not observers. Operations are based on
attempts (you can specify time between them) where observers are based on
patience and operation latency SLA. (you can specify frequency, but
conceptually leave it to the backend to determine so it could use different
strategies in the future).

jtk54:
Ok, makes sense. LGTM

PR #213, opened 2016-09-26
Added DICT_MATCHES predicate

ewiseblatt:
@jtk54


citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
                            path_value=PathValue('', value), valid=valid)
 
 
+class DictMatchesPredicate(BinaryPredicate):
+  """Implements binary predicate comparison predicates against dict values.
+
+  The dictionary value is in the form {field : pred} where each element
```
jtk54: Can `field` be an arbitrary path?

citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
...
+  def strict(self):
+    """Whether only the specified fields may be present (True) or not."""
+    return self.__strict
+
+  def __init__(self, operand, **kwargs):
+    """Constructor."""
+    if not isinstance(operand, dict):
+      raise TypeError(
+          '{0} is not a dict: {1!r}'.format(operand.__class__, operand))
+    self.__strict = kwargs.pop('strict', False)
```
jtk54: Might want `strict` as a separate named parameter here.

citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
...
+  """
+
+  @property
+  def strict(self):
+    """Whether only the specified fields may be present (True) or not."""
+    return self.__strict
+
+  def __init__(self, operand, **kwargs):
+    """Constructor."""
+    if not isinstance(operand, dict):
```
jtk54: `operand` is the json payload returned in the call? Where does this come from?

citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
                            path_value=PathValue('', value), valid=valid)
 
 
+class DictMatchesPredicate(BinaryPredicate):
+  """Implements binary predicate comparison predicates against dict values.
+
+  The dictionary value is in the form {field : pred} where each element
+  in the dictionary is a predicate for validating that particular field.
+  A strict predicate means that exactly all specified fields must be present.
+  Otherwise this permits additional fields with arbitrary values.
```
jtk54: Might be worth including a simple example in the docstring.

citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
...
+      raise TypeError(
+          '{0} is not a dict: {1!r}'.format(operand.__class__, operand))
+    self.__strict = kwargs.pop('strict', False)
+    super(DictMatchesPredicate, self).__init__('Matches', operand, **kwargs)
+
+  def __call__(self, context, value):
+    """Implements Predicate interface.
+
+    context: [ExecutionContext] The execution context.
+    value: [dict] The dictionary to match against our operand.
```
jtk54: Maybe something like `expected_value`?

citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
...
+    valid = True
+    # pylint: disable=redefined-variable-type
+    for name, pred in self.operand.items():
+      name_value = value.get(name)
+      name_result = None
+      if name_value is None:
+        name_result = MissingPathError(value, name)
+      else:
+        path_result_builder = PathPredicateResultBuilder(
+            source=value, pred=pred)
```
jtk54: Is this meant for logging/diff purposes? Seems a bit weird to store the whole dictionary for a predicate picked out by name.

citest/json_predicate/binary_predicate.py
```diff
@@ -146,6 +150,83 @@ def __call__(self, context, value):
...
+            PathValue(name, name_value), pred(context, value.get(name)))
+        name_result = path_result_builder.build()
+
+      if not name_result:
+        valid = False
+      match_result_builder.append_result(name_result)
+
+    if self.strict:
+      # Only consider and add strictness result if it fails
+      strict_result = self._find_unexpected_path_errors(context, value)
```
jtk54: `strict_result_errors` is more clear.

citest/json_predicate/path_result.py
```diff
@@ -181,6 +181,28 @@ def __init__(self, source, target_path, path_value=None, **kwargs):
...
 
 
+class UnexpectedPathError(PathResult):
+  """A PathResult indicating an object contained an unexpected attribute."""
+
+  def __init__(self, source, target_path, path_value, **kwargs):
+    """Constructor.
+
+    Args:
+      source: [obj] The original JSON object path_value is relative to.
```
jtk54: `obj` or `dict`? I think by default JSON objects are represented as `dict`s in python.

tests/json_predicate/binary_predicate_test.py
```diff
@@ -195,6 +195,168 @@ def test_dict_ne(self):
...
+                  .build(True))
+              .build(True))
+
+    self.assertTrue(result)
+    self.assertEquals(expect, result)
+
+  def test_dict_match_simple_bad(self):
+    context = ExecutionContext()
+    source = {'n' : 10}
+    want = {'n' : jp.NUM_NE(10)}
```
jtk54: I like that this is moving toward simpler table/map based tests. This feels more intuitive and cleaner than chaining a bunch of predicates together.

tests/json_predicate/binary_predicate_test.py
```diff
@@ -195,6 +195,168 @@ def test_dict_ne(self):
...
+                      jp.PathValue('n', 10), jp.NUM_NE(10)(context, 10))
+                  .build(False))
+              .build(False))
+
+    self.assertFalse(result)
+    self.assertEquals(expect, result)
+
+  def test_dict_match_multi_ok(self):
+    context = ExecutionContext()
+    source = {'a' : 'testing', 'n' : 10}
```
jtk54: In a normal integration test scenario, would `source` be the result coming back from something like a `list_resource`? How would that look connected into `DICT_MATCHES`.

citest/json_predicate/binary_predicate.py
jtk54: Also, please add that to the docstring.

citest/json_predicate/binary_predicate.py
ewiseblatt: This is the actual value. The operand is the expected value. All the predicates have "value" as the actual value. All the binary predicates have "operand" as the expected value, which isnt really "expected value" since it is a reference point (e.g. LT or SUBTRING arent expected values). Consistent naming is important if you want to pass as keywords rather than positional parameters where sometimes it might not be clear what's what.

citest/json_predicate/binary_predicate.py
ewiseblatt: Maybe. I'm not doing that, but I dont have hard and fast rules for when I do and dont.
I dont expect it to be common to use and wanted to keep the arguments lists small.
When It is present, it is standard this way. So this is consistent with other uses,
though as you point out, might not be the ideal standard pattern.

citest/json_predicate/binary_predicate.py
ewiseblatt: All the BinaryPredicates are constructed with an operand, which is the "operand" in the binary operator that you are comparing against. The other operand is passed into the __call__ method as the value when it comes time to test a particular value.

citest/json_predicate/binary_predicate.py
ewiseblatt: I'd like to, but not in this PR. For now you'd have to use path predicates for sparse trees. I think the cases I've seen so far were only one or two levels deep, so thought I could get away without it for now.

citest/json_predicate/binary_predicate.py
ewiseblatt: Every PredicateResult derived from a value (as opposed to composition) has a source, which is the instance value that led the the result. The value here would be the value from the key value pair. So it isnt the original source dictionary, it is the value for the attribute name that maps to the predicate specified for that attribute. I think it is what you want.

tests/json_predicate/binary_predicate_test.py
ewiseblatt: I havent run this through the reporting yet, but think it will yield clearer reports as well.

tests/json_predicate/binary_predicate_test.py
ewiseblatt: DICT_MATCHES would be looking for an element of the list that has the properties specified by the operand.

ewiseblatt: More precisely, the verifier would iterate over the list and apply the DICT_MATCHES to each of the elements it finds, or until it finds one that matches. Depeneds on what the semantics of the verifier is as to how eager it is.

citest/json_predicate/binary_predicate.py
jtk54: If it's an actual value then, can we go with `actual_value`?

citest/json_predicate/binary_predicate.py
jtk54: Makes sense.

citest/json_predicate/binary_predicate.py
ewiseblatt: what's wrong with value. Passing a value into the __call__ method suggests it being an actual value. This method is part of the interface.
https://github.com/google/citest/blob/master/citest/json_predicate/predicate.py#L33

jtk54: Ok makes sense.

jtk54:
This LGTM. I'd like to see what the journal output looks like using this to create the operation contracts.

ewiseblatt:
Here's the additional changes to produce the reporting output I showed you offline.

PR #215, opened 2016-09-27
Resolved naming collisions between http upsert tests.

jtk54:
@lwander please review. These tests share a scenario, so they were performing illegal operations on components when run in parallel in the nightly builds.


ewiseblatt:
could you hold off on this?
I moved this part of the repo to spinnaker/spinnaker but am awaiting approval.
So the change would need to happen in spinnaker/spinnaker/testing/citest once it goes there.
looks ok otherwise.

jtk54:
The Jenkins build will fail until this fix is in.

ewiseblatt:
Ok. Put it in. I'll re-import or copy over the current files.

PR #218, opened 2016-09-28
Http LB tests integrated into Google smoke tests.

jtk54:
Wires two simple Http LB cases into the Google smoke tests. @ewiseblatt please review.

### DIFF
--- a/spinnaker/spinnaker_system/google_smoke_test.py
+++ b/spinnaker/spinnaker_system/google_smoke_test.py
@@ -51,12 +51,14 @@ import citest.json_predicate as jp
 import citest.service_testing as st
 
 # Spinnaker modules.
+from google_http_lb_upsert_scenario import GoogleHttpLoadBalancerTestScenario
+
 import spinnaker_testing as sk
 import spinnaker_testing.gate as gate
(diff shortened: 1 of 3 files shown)


spinnaker/spinnaker_system/google_smoke_test.py
```diff
@@ -51,12 +51,14 @@
...
 # Spinnaker modules.
+from google_http_lb_upsert_scenario import GoogleHttpLoadBalancerTestScenario
+
 import spinnaker_testing as sk
 import spinnaker_testing.gate as gate
 import citest.base
 
 
-class GoogleSmokeTestScenario(sk.SpinnakerTestScenario):
+class GoogleSmokeTestScenario(GoogleHttpLoadBalancerTestScenario):
```
ewiseblatt: I would rather you didnt do this inheritance.
Could you do something like GoogleHttpLoadBalancerTestScenario(self.scenario.bindings, self.scenario.agent).upsert_min_load_balancer())

jtk54:
@ewiseblatt removed the inheritance. It's ugly but works.

PR #220, opened 2016-10-01
Rewrote some existing contracts to use DICT_MATCHES and LIST_MATCHES.

ewiseblatt:
rather than ANDing together individual paths. I think this is easier
to understand, and reports a little better. Syntax is still not ideal,
and dealing with lists vs elements is still more troublesome than I wish,
but this is no worse in that regard.

Also, seeing errors with AWS that look like the cli might be returning different errors than before (either reworded them or failing differently) so I updated the AWS error handling to accomodate this new reality.

@jtk54


README.md
```diff
@@ -55,24 +55,25 @@ A test case might look something like this:
...
                                 retryable_for_secs=30)
      .list_resources('target-pools')
      .contains_path_value('name', '%s-tp' % load_balancer_name))
     (builder.new_clause_builder('Forwarding Rules Added',
                                 retryable_for_secs=30)
      .list_resources('forwarding-rules')
-     .contains_pred_list([
-          json_predicate.PathContainsPredicate('name', load_balancer_name),
-          json_predicate.PathContainsPredicate('target', target_pool_name)]))
+     .contains_match({
```
jtk54: I'm not sure `contains_match` makes a lot of sense for what this function does, which is validating predicates based on attribute paths.

overview.md
```diff
@@ -395,11 +401,17 @@ that it used to ensure uniqueness. However we know part of the substring that
 the name should have (and have ourselves chosen it to be unique).
 
 The other clauses are simlilar but happen to use different types of predicates.
-The `contains_pred_list()` is checking that each of the predicates holds on a
-particular object. This would be the same as the `jp.DICT_SUBSET` predicate,
-however reporting is different should one of these predicates fail, the individual
-attributes will be shown as opposed to the entire dict since these are independent
-predicates.
+The `contains_match()` is a shorthand notation for specifying a collection of
```
jtk54: See above about naming.

overview.md
```diff
@@ -395,11 +401,17 @@ that it used to ensure uniqueness. However we know part of the substring that
...
+object attributes with their validation predicates, all of which must be
+satisifed by the same object. This is functionally equivalent to listing all
+the individual path predicates, but more concise to read and write, and allows
+more concise reporting as well.
+
+Note that the `path_contains_value(path, {key: value})` call in the previous
```
jtk54: I think we should work out how to get the same reporting while using the more complex path keys. Nested dictionaries are hard to reason about -- I'd rather have a flat dictionary with paths pointing to predicates.

spinnaker/spinnaker_system/aws_smoke_test.py
```diff
@@ -236,16 +236,13 @@ def upsert_load_balancer(self, use_vpc):
...
-         jp.PathPredicate(
-             'LoadBalancerDescriptions/AvailabilityZones{0}'.format(
-                 jp.DONT_ENUMERATE_TERMINAL),
-             jp.LIST_SIMILAR(expect_avail_zones)),
-         jp.PathElementsContainPredicate(
-             'LoadBalancerDescriptions/ListenerDescriptions', listener)
-         ])
+     .contains_path_match(
+          'LoadBalancerDescriptions',
+          {'HealthCheck': jp.DICT_MATCHES(
```
jtk54: Having a single layer of keys to predicates would be much easier to read here.

spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
```diff
@@ -181,62 +181,63 @@ def _add_contract_clauses(self, contract_builder, upsert):
...
+      )
     for hc in health_checks:
-      hc_clause_builder.contains_pred_list(
-        [
-          jp.PathEqPredicate('name', hc['name']),
-          jp.PathEqPredicate('requestPath', hc['requestPath']),
-          jp.PathEqPredicate('port', hc['port'])
-        ]
+      hc_clause_builder.contains_match({
+          'name' : jp.STR_EQ(hc['name']),
```
jtk54: I would eventually opt for more verbose names than `STR_EQ`, etc. `STR_EQ` is a trivial case, but I can imagine users of this being confused and continually looking up what these mean.

spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
```diff
@@ -181,62 +181,63 @@ def _add_contract_clauses(self, contract_builder, upsert):
...
+        path_rules_specs.append({
+            'service': jp.STR_SUBSTR(pr['backendService']['name']),
+            'paths': jp.LIST_MATCHES([jp.STR_EQ(path) for path in pr['paths']])
+            })
+
+      url_map_clause_builder.contains_match({
+         'name': jp.STR_EQ(self.__lb_name),
+         'defaultService':
+             jp.STR_EQ(self._get_bs_link(upsert['defaultService']['name'])),
+         'pathMatchers': jp.LIST_MATCHES([
```
jtk54: This is still pretty hard to read even with one level of nesting.

README.md
ewiseblatt: It is checking whether the resource list contains an object matching the specified prototype. What were you thinking?

overview.md
ewiseblatt: This is an incremental step. I dont think nested dictionaries are hard to reason about, but they are harder to read and write and less concise overall.

spinnaker/spinnaker_system/aws_smoke_test.py
ewiseblatt: Sure, but this is an incremental step. I think it is better than using contains_pred_list. My priority was to migrate the registry over into the spinnaker repository and wanted to point in this direction rather than using contains_pred_list so I can get rid of that and use this style of predicate ("name -> predicate" as opposed to "name->operand") to be more consistent with using the more concisely expressive notation.

spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
ewiseblatt: I could say jp.EQUIVALENT. I thought I'd be more explicit about what is being compared.

spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
ewiseblatt: Agreed. This is what I was refering to in the PR note when I said expressing lists vs elements is still awkward.

PR #224, opened 2016-10-05
Updated L7 upsert description for recent refactor.

jtk54:
@ewiseblatt This takes care of the recent L7 test failures.


ewiseblatt:
oh, if this is a test failure then go ahead and submit.

jtk54:
Yep test failure -- submitting.

PR #226, opened 2016-10-05
L7 upsert test: Make 'add server group' use tags, not security group.

jtk54:
@ewiseblatt please review. Creating a server group previously used the `securityGroups` field to specify security groups, but now only uses `tags`. I modified the operation descriptions for creating the security group and server group to use the tags. This should fix the previous test failure.


spinnaker/spinnaker_system/google_http_lb_upsert_scenario.py
```diff
@@ -659,6 +660,7 @@ def delete_server_group(self):
       job=[{
         'cloudProvider': 'gce',
         'serverGroupName': group_name,
+        'asgName': group_name,
```
jtk54: That's not relevant, I'll take it out. I just made it match what's coming from Deck.

jtk54:
Relatively new. It's actually quite unclear why some of the tests were succeeding earlier; we were adding the server group with incorrect security groups the whole time. I'll investigate that after this issue is resolved.

ewiseblatt:
LGTM

PR #227, opened 2016-10-07
Added rendering transforms to simplify report structure

ewiseblatt:
@jtk54 

The data model dumps show the literal data model. There are some
common types that use lists for a general case so these are reported
out as lists even if the particular instance is only dealing with one
element.

If the list is just an implementation detail (as opposed to something
that is fundamentally a list and may contain one) then this list rendering
adds abstraction overhead and complexity for the reader. This PR allows
certain types to explicity transform the interpretation of their encodings
depending on the context of the instance value.

For example PathValueResult contains a list of values at a given path
because in general paths can pass through lists as intermediate nodes,
so have different branches down the path resulting in different values.
If there is no list, then the PathValue is a list of 1, but that is just
an implementation detail. Conceptually  the path value is the element in
that containing list so should be reported that way.

The transforms are on the reporting side rather than on the snapshot side.
This might be a mistake, but I think that there could be other reasons to
read snapshots in the future so having a consistent data model will be
simpler. The transforms on the interpretation are not tied to the HTML
rendering, so all renderers (and readers of journal data for that matter)
could share the same implementation as if it were in the snapshot anyway.

### DIFF
--- a/citest/reporting/simplify_entity_transforms.py
+++ b/citest/reporting/simplify_entity_transforms.py
@@ -1,0 +1,264 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helper classes and methods for special case interpreting Snapshot Entities.
+
+These classes rewrite some of the entities to simplify their reporting
+interpretations to provide a more concise data model.
+"""
+
+class EdgeLabelValueTransformer(object):
+  """Transforms a snapshot entity edge label and value depending on context.
+
+
+  The instance is bound to the original entity and is called with each edge
+  to extract the label and value for that edge giving the instance a chance
+  to rewrite the label and/or value, or to ignore it completely.
+  """
+
+  @property
+  def entity(self):
+    """The SnapshotEntity dictionary bound at construction."""
+    return self.__entity
+
+  @property
+  def entity_manager(self):
+    """The EntityManager bound at construction."""
+    return self.__entity_manager
+
+  @staticmethod
+  def _find_edge(entity, label_name):
+    """Return the edge with the given label name.
+
+    Args:
+      entity: [dict] The snapshot entity dict
+      label_name: [string] The label name we're looking for.
+    """
+    if entity is None:
+      return None
+    for edge in entity.get('_edges', []):
+      if edge.get('label') == label_name:
+        return edge
+    return None
+
+  def __init__(self, entity, entity_manager):
+    """Constructor."""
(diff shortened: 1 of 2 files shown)


citest/reporting/simplify_entity_transforms.py
```diff
@@ -0,0 +1,264 @@
...
+    """The SnapshotEntity dictionary bound at construction."""
+    return self.__entity
+
+  @property
+  def entity_manager(self):
+    """The EntityManager bound at construction."""
+    return self.__entity_manager
+
+  @staticmethod
+  def _find_edge(entity, name):
```
jtk54: s/`name`/`label_name`.

citest/reporting/simplify_entity_transforms.py
```diff
@@ -0,0 +1,264 @@
...
+  def __call__(self, edge):
+    """Returns label, value for the given edge.
+
+    Args:
+      edge: [dict] The edge dictionary
+
+    Returns:
+      label, value tuple where label is None means ignore the edge.
+    """
+    label = edge.get('label', '?unlabeled')
```
jtk54: What are the issues with making the default `None`?

citest/reporting/simplify_entity_transforms.py
```diff
@@ -0,0 +1,264 @@
...
+    """Constructor."""
+    super(SequencedResultLabelValueTransformer, self).__init__(
+        'Results', 'Result', entity, entity_manager)
+
+  def __call__(self, edge):
+    """Implements EdgeLabelValueTransformer."""
+    if self.has_exactly_one:
+      # Drop the count attribute.
+      label = edge.get('label')
+      if label == '#':
```
jtk54: Just out of curiosity, are the `export_to_snapshot` functions documented with the things they encode here? These seem (rightly) to be different for each Predicate type, but they should definitely be documented.

ewiseblatt:
I took a journal and rendered it both the old way and the new way and show
a similar vignette below.
In the first of the green results, I opened up one level of Path Value and
Justifications, which are one element lists.
In the red results there is a map over a single object which provides
another example


Note above that opening the path values /justification reveals a list
below, but
there is only one element in the list. It's just a value not really a list
of values so this
is annoying and adds extra clutter even when opened, where as below it
opens directly
into the values because the list containing a value is transformed into
just the value itself.

In the red part note the [0] column between "Results" and the actual
result. This goes away
below. SImilar principle to above. The [0], [1] values remain because the
list is in fact a list.
Also note that below the "# 1" attribute row went away entirely because the
attribute was
filtered out along with the list being transformed.

citest/reporting/simplify_entity_transforms.py
ewiseblatt: Edges should always have a label to name the relationship to the value. However this isnt strictly enforced in the writer, and this library is decoupled from the data anyway so I need to accomodate errors. Giving it a visible name hints the fact that it wasnt named, which should be addressed. This much isnt a change, I just moved it from being hardcoded in html_renderer.py to being the default here when html_renderer runs through the transform. What's new is that I now use None to mean ignore the value altoghether. Since the source cannot contain None (we give None values names here), this is unambiguous.

citest/reporting/simplify_entity_transforms.py
ewiseblatt: The journal documents some things, but I havent formalized everything yet. I tried not to make any specific for predicate types, though not all predicates use each. Some motivated introducing something, but I tried to tie them to reporting (or data model) concepts as opposed to predicate implementations.

PR #228, opened 2016-10-07
Refactored DICT_MATCH so that it allows multi-level path attributes.

ewiseblatt:
@jtk54 

I also changed PathPredicate to allow configuration of whether or not
to expand list terminals. It defaults to exapand but DICT_MATCh makes
it not. I think this makes it more natural to specify constraints on
attributes, and the feature could be used for PathPredicate cases.

### DIFF
--- a/citest/json_predicate/map_predicate.py
+++ b/citest/json_predicate/map_predicate.py
@@ -141,6 +141,7 @@ class MapPredicateResult(SequencedPredicateResult):
 
   def __init__(self, valid, pred, obj_list, all_results,
                good_map, bad_map, **kwargs):
+    # pylint: disable=too-many-arguments
     self.__obj_list = obj_list
     self.__good_map = good_map
     self.__bad_map = bad_map
@@ -152,6 +153,21 @@ class MapPredicateResult(SequencedPredicateResult):
             and self.__obj_list == result.obj_list
             and self.__good_map == result.good_object_result_mappings
             and self.__bad_map == result.bad_object_result_mappings)
+
+  def clone_with_source(self, source, base_target_path, base_value_path):
+    """Implements CloneableWithNewSource interface."""
+    builder = MapPredicateResultBuilder(self.pred)
+
+    for index, orig in enumerate(self.results):
+      if isinstance(orig, predicate.CloneableWithNewSource):
+        result = orig.clone_with_source(source=source,
+                                        base_target_path=base_target_path,
+                                        base_value_path=base_value_path)
+      else:
+        result = orig
+      builder.add_result(self.__obj_list[index], result)
+    return builder.build(self.valid)
+
 
 class MapPredicate(predicate.ValuePredicate):
   """Applies a predicate to all elements of a list or a non-list object.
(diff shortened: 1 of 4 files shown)


citest/json_predicate/map_predicate.py
```diff
@@ -153,6 +153,21 @@ def __eq__(self, result):
             and self.__good_map == result.good_object_result_mappings
             and self.__bad_map == result.bad_object_result_mappings)
 
+  def clone_with_source(self, source, base_target_path, base_value_path):
+    """Implements CloneableWithNewSource interface."""
+    builder = MapPredicateResultBuilder(self.pred)
+
+    for index, orig in enumerate(self.results):
+      if isinstance(orig, predicate.CloneableWithNewSource):
```
jtk54: Does python consider an one object an instance of another if it just overrides a function? It seems like there's no explicit object dependency that would cause this to be true.

jtk54: For instance, in Java there's `implements ...` where you expect certain functions to be present. Is this Python's version of the same thing?

citest/json_predicate/path_predicate.py
```diff
@@ -292,8 +315,8 @@ def __call__(self, context, source):
 
     path = context.eval(self.__path)
 
-    builder = PathPredicateResultBuilder(pred=self, source=source)
-    enumerate_terminal = True
+    builder = PathPredicateResultBuilder(pred=self.source_pred, source=source)
```
jtk54: If `source_pred` is `None`, self is used here, correct?

tests/json_predicate/binary_predicate_test.py
```diff
@@ -302,21 +302,21 @@ def test_list_match_strict_bad(self):
     self.assertFalse(result)
     self.assertEquals(expect, result)
 
+  def _match_dict_attribute_result(self, context, pred, key, value):
+    return jp.PathPredicate(key, pred, source_pred=pred,
```
jtk54: Seems weird we're passing in `pred` to two parameters here...

citest/json_predicate/map_predicate.py
ewiseblatt: Often in python you'd take advantage of duck typing. But I'm using isinstance because I want to indicate class semantics, which I think is appropriate here. The framework has a few marker classes (e.g. "CloneableWithNewSource", "Snapshotable") the guide key heuristics.

CloneableWithNewSource is defined here: https://github.com/google/citest/blob/776b27277dfc74204c5ca649c9f2cde0c62ae0b6/citest/json_predicate/predicate.py#L147

Python classes support multiple inheritence. The base class(es) are specified with the class declaration ("object" being the root class). Here's an example of a predicate result saying that it is CloneableWithNewSource.
https://github.com/google/citest/blob/e85bb27b3bcf1bd2e158c2be1eac0b8964dcf807/citest/json_predicate/path_result.py#L29

citest/json_predicate/path_predicate.py
ewiseblatt: #288 prevents source_pred from ever being None. It will default to self. Actually, if you explicitly set it to None then it will be None and result in an error. I can change #288 or be "or self" instead of defaulting to self.

tests/json_predicate/binary_predicate_test.py
ewiseblatt: The above goes for the implementation. This test is a whitebox test that knows that's the result that should be returned.

PR #231, opened 2016-10-22
Refactored HTML rendering to use BeautifulSoup API

ewiseblatt:
@jtk54 

This is a refactoring of the HTML generation to use beautiful soup as you suggested earlier last week. Except it doesnt.

While it is cleaner, it took >10x longer to run. That's especially surprising because I'm not really parsing HTML -- I am generating it. So that means the overhead is likely in building and managing the data model. To process the http load balancer test journal went from 12s to over 2m. That's extremely significant and a turn off from wanting to do this.

Another interesting stat is that I turned off formatting which causes the size to drop from 173M to 44M yet the time to write it out increases from around 1s to 33s. The BS write is doing to conversion from its data model into HTML which is why it takes so long, yet even so it is almost 3x as long as my original implementation took to convert from the journal data model into HTML, which is ultimately the same logical data model.

Out of curiousity, I wrote my own minimal implementation of the BeautifulSoup API from only the calls I make. I produce the same file in a little over 30s of which around 9s is converting from my API into HTML (without control over formatting). So it is about 2.5x longer than the original, but 4x faster than using Beautiful Soup.

Since the implementation is pretty small and simple, I decided that I'd use that instead of BeautifulSoup and get some improvement without the full cost and make it easily swappable later if tradeoff values change.

### DIFF
--- a/citest/reporting/html_document_manager.py
+++ b/citest/reporting/html_document_manager.py
@@ -22,10 +22,112 @@ The renderer module handles more of the data encoding and type specific things.
 import cgi
 
 
+class MyTag(object):
+  """Denotes an HTML tag and its content.
+
+  This interface is a minimal subset of BeautifulSoup that we use on tags.
+  The class is a more efficient implementation that runs in 25% of the time
+  (but still about 2.5x longer than hand-constructing HTML).
+  """
+  # pylint: disable=too-few-public-methods
+
+  def __init__(self, name, **kwargs):
+    """Construct a tag
+
+    Args:
+      name: [string] The tag name
+      kwargs: [kwargs] Optional tag key/value attributes
+    """
+    self.__name = name
+    self.__attrs = dict(kwargs)
+    self.__parts = []
+    self.string = ''
+
+  def __setitem__(self, key, value):
+    """Add additional key/value attributes."""
+    self.__attrs[key] = value
+
+  def append(self, part):
+    """Add additional contents (children) to a tag.
+
+    Args:
+      part: [MyTag/String/DocFragment] The thing to add.
+    """
+    self.__parts.append(part)
+
+  def __str__(self):
+    attributes = [' {key}="{value}"'.format(key=key, value=value)
+                  for key, value in self.__attrs.items()]
+    return '<{name}{attributes}>{text}{parts}</{name}>'.format(
+        name=self.__name,
+        attributes=''.join(attributes),
+        text=self.string,
+        parts=''.join([str(part) for part in self.__parts]))
+
+
+class MyDocFragment(object):
+  """Denotes a block of HTML text.
+
+  This interface is a minimal subset of BeautifulSoup that we use.
+  Note that unlike BeautifulSoup, we wont parse raw HTML we are given.
+  Therefore we wont be able to navigate or edit the internal structure.
+  However, since we are constructing the document, we shouldnt need to.
+
+  The class is a more efficient implementation that runs in 25% of the time
+  (but still about 2.5x longer than hand-constructing HTML).
+  """
+  def __init__(self, html):
+    """Construct fragment from initial HTML.
+
+    Args:
+      html: [string] Escaped HTML text that starts the fragment.
+    """
+    self.__parts = [html]
+
+  def append(self, part):
+    """Add a block to the end of the current fragment.
+
+    Args:
+      part: [any]  Part should be a Tag/String/DocFragment
+    """
+    self.__parts.append(part)
+
+  def __str__(self):
+    return ''.join([str(part) for part in self.__parts])
+
+  def new_tag(self, name, **kwargs):
+    """Create a new tag.
+
+    Args:
+      name: [string] The tag name.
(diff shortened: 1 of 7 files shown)


citest/reporting/html_document_manager.py
```diff
@@ -223,96 +391,86 @@ def make_expandable_tag_attr_pair(self, section_id, default_expanded):
...
-      <td class="input">The attribute denotes an input data value, or an object that acted as an input.</td>
-  <tr><th class="output">Output</th>
-      <td class="output">The attribute denotes an output data value, or an object that acted as an output.</td>
-  <tr><th class="control">Control</th>
-      <td class="control">The attribute denotes a control value used to configure some related component.</td>
-  <tr><th class="mechanism">Mechanism</th>
-      <td class="mechanism">The attribute denotes a component used as a mechanism providing behaviors to another component.</td>
-</table>
-"""
+    add_row = lambda klass, name, help: (self.make_tag_container(
```
jtk54: A function like this could be useful in other places.

citest/reporting/html_document_manager.py
```diff
@@ -223,96 +391,86 @@ def make_expandable_tag_attr_pair(self, section_id, default_expanded):
...
-        '{indent}</table>\n'.format(indent=indent)]
-    return '\n'.join(lines)
-
-  def build_html_head_block(self, title):
-    """Builds text for the HTML HEAD section.
-
-    The HEAD section will contain the standard Javascript and CSS used within.
+    expanded_tag_attrs_dict, hidden_tag_attrs_dict = (
+        self.make_expandable_tag_attr_kwargs_pair(
+            section_id=section_id, default_expanded=False))
```
jtk54: Why do you use these as named parameters instead of positional?

citest/reporting/html_document_manager.py
```diff
@@ -223,96 +391,86 @@ def make_expandable_tag_attr_pair(self, section_id, default_expanded):
...
+    td_tag.append(expand_tag)
+    td_tag.append(hide_tag)
 
-  def build_end_html_document(self):
-    """Writes the closing of the HTML document BODY and HTML tags."""
-    return '</body>\n</html>'
+    table_tag = self.new_tag('table')
+    tr_tag = self.new_tag('tr')
+    tr_tag.append(td_tag)
+    table_tag.append(tr_tag)
```
jtk54: See the `append_new_tag` comment upstairs. Would shorten this a bit.

citest/reporting/html_renderer.py
```diff
@@ -362,30 +282,24 @@ def process_metadata(self, obj, blacklist=[]):
...
-    formatter = self.__formatter
-    table_indent = formatter.line_indent
-    fragments = ['<table>\n']
-    formatter.push_level()
-    line_indent = formatter.line_indent
+    table = self.__document_manager.new_tag('table')
     row_count = 0
     for name, value in obj.items():
-      if name not in blacklist:
+      if name not in blacklist or []:
```
jtk54: I believe `[]` is falsy in python, so this is unneeded. Do you mean `blacklist != []`?

citest/reporting/html_document_manager.py
```diff
@@ -22,10 +22,112 @@
...
+
+    Args:
+      part: [MyTag/String/DocFragment] The thing to add.
+    """
+    self.__parts.append(part)
+
+  def __str__(self):
+    attributes = [' {key}="{value}"'.format(key=key, value=value)
+                  for key, value in self.__attrs.items()]
+    return '<{name}{attributes}>{text}{parts}</{name}>'.format(
```
ewiseblatt: That's what the leading space in the string is for. Doing a ' '.join would only put a space between multiple attributes, not before the initial one. The way I have it avoids special casing and gives you exactly the spaces you need.

citest/reporting/html_document_manager.py
```diff
@@ -22,10 +22,112 @@
...
+
+    Args:
+      part: [any]  Part shuld be a Tag/String/DocFragment
+    """
+    self.__parts.append(part)
+
+  def __str__(self):
+    return ''.join([str(part) for part in self.__parts])
+
+  def new_tag(self, name, **kwargs):
```
ewiseblatt: MyTag is an internal implementation detail. new_tag is the BS API. I implement new_tag by delegating to the MyTag class.

citest/reporting/html_renderer.py
ewiseblatt: "or []" is for when blacklist=None because "not in" wouldnt be valid on None.

citest/reporting/html_document_manager.py
jtk54: Yep makes sense.

jtk54:
LGTM

citest/reporting/html_document_manager.py
jtk54: Alright, makes sense if we do need the HTML parsing/searching. I was thinking from a perspective where we are just building HTML, possibly using a different library to do the same thing.

PR #234, opened 2016-10-24
Can exclude resources.

ewiseblatt:
I want this to avoid deleting "operations".
Also, dont bother listing things that cannot be deleted. They appear to be things like zones.

@lwander


ewiseblatt:
ping

lwander:
LGTM

PR #235, opened 2016-10-26
Refactored displaying diff output.

ewiseblatt:
@jtk54 
Sorry to have to do this. It's basically adding a data structure and shuffling code around. The heuristics and stuff are the same. It lets me store what's different rather than just print it, even though in the end I just print it. The value here is that I can clean up the output.


citest/gcp_testing/resource_snapshot.py
```diff
@@ -303,6 +317,142 @@ def stringify(self, resource):
     return '\n'.join(lines)
 
 
+class ApiDiff(object):
```
jtk54: Might be worth collecting the api family diffs in another class and using that to clean up the representation.

PR #236, opened 2016-10-29
Added simple example

ewiseblatt:
Some cleanups and bug fixes.

Let contract "match" methods take either a list to indicate a LIST_MATCH
in addition to the existing semantics where a dict indicates a DICT_MATCH.

@jtk54 
The example helps documents how to use various features when writing tests.


examples/keystore/async_keystore_test.py
```diff
@@ -0,0 +1,299 @@
...
+
+    parts = http_response.output.split(' ', 1)
+    self.__current_state = parts[0]
+    self.__error = parts[1] if len(parts) > 1 else None
+
+
+class AsynchronousKeystoreTestScenario(st.AgentTestScenario):
+  """Example test scenario. This is not that interesting.
+
+  It is not actually used, but we could use it to coordinate testing values
```
jtk54: Might be worth it to mention that the pattern the rest of the tests use is to define test cases in the scenario and use them in the test. Seems like we reversed that pattern a bit here.

ewiseblatt: Yea, maybe I should switch this back to use the scenario.

PR #237, opened 2016-11-03
Added ConfigurationBindings management.

ewiseblatt:
Use ConfigurationBindings for the bindings instead of the argparser.
Deprecated initArgumentParser in favor of init_bindings_builder.
The ConfigurationBindingsBuilder can also create command-line arguments
so that bindings can be replaced either by command-line or config files.

With the deprecation, this is fully backward compatible with the existing
tests. I'm going to migrate the tests over in another PR and give other
authors a chance to migrate before I take out the initArgumentParser api.

@jtk54


citest/base/bindings.py
```diff
@@ -0,0 +1,376 @@
...
+          package_paths.append(package_path)
+
+      self.__add_configs_for_name(parts[0], location, [], found_list)
+      self.__add_configs_for_name(package, None, dirs, found_list)
+      package = '.'.join(parts[:-1]) if len(parts) > 1 else None
+      location = os.path.dirname(location)
+    found_list.extend(package_paths)
+
+    # Add in reverse order so highest precedence is last
+    self.__config_files.extend(found_list[::-1])
```
jtk54: Clarify what this means for the config files in the directory structure. Further down in the tree means lower priority?

citest/base/test_runner.py
```diff
@@ -204,6 +215,14 @@ def set_parser_inits(self, inits):
     """
     self.__parser_inits = inits or []
 
+  def set_config_files(self, config_files):
+    """Provides a means for setting the config_files attribute.
+
+    This is intentionally not an assignment because it is not intended to be
+    called, but is here in case it is no possible to use the "main()" method.
```
jtk54: s/no possible/not possible.

citest/base/bindings.py
```diff
@@ -0,0 +1,376 @@
...
+  of python "ConfigParser" to specify them in a static file.
+
+  The search space is (in order of precedence):
+    Overrides - These are explicit values injected at runtime.
+    ConfigParser - These are values read from configuration files[*].
+    Defaults - These are static default values if not found in a configuration
+       file.
+
+
+  The ConfigParser has an additional hierarchy of sections. If the instance
```
ewiseblatt: Yes but not the literal "instance" rather the value of self.__section.

citest/base/bindings.py
```diff
@@ -0,0 +1,376 @@
...
+  def overrides(self):
+    """Returns the underlying overrides dictionary."""
+    return self.__overrides
+
+  def __init__(self, config_parser, overrides, section=None):
+    """Construct a bindings instance.
+
+    Args:
+      config_parser: [ConfigParser] Retrieves values from configuration files.
+      overrides: [dict] Values that override anything found in config_parser.
```
ewiseblatt: Updated the Overrides section in the class docstring. Flags are overrides, but there could be other mechanisms for this as well, like an explicit assignment.

citest/base/bindings.py
```diff
@@ -0,0 +1,376 @@
...
+
+    for section in all_sections:
+      if self.__config_parser.has_option(section, key):
+        return True
+    return False
+
+  def __getitem__(self, name):
+    """Retrieve a binding value."""
+    have = self._do_get(name, None)
+    if not self.__contains__(name):
```
ewiseblatt: This meant to check "have is not None" before bothering with contains.

citest/base/bindings.py
```diff
@@ -0,0 +1,376 @@
...
+    flags.update({key.lower(): value
+                  for key, value in self.__overrides.items()})
+
+    bindings = ConfigurationBindings(
+        config_parser, overrides=flags, section=section)
+    return bindings
+
+  def _exists(self, path):
+    """Helper function to determine if a path exists to facilitate testing."""
+    return os.path.exists(path)
```
ewiseblatt: It's easy to intercept to test what paths are being considered.
I didnt write the test yet. Turns out there is a trick because the test doesnt always run as a module so it's behavior changes depending on how it is run so not sure how to actually write the test. I verified it manually for the time being.

PR #239, opened 2016-11-08
Suppress warning from google apiclient.

ewiseblatt:
@ttomsu 
This works around the problem you were seeing last week.
I dont like it, but dont see a way to force every different script to update.


ewiseblatt:
ping

ttomsu:
LGTM

PR #241, opened 2017-01-03
Added match_kwargs for verifying matches

ewiseblatt:
@jtk54


citest/json_contract/value_observation_verifier.py
```diff
@@ -80,22 +80,25 @@ def add_constraint(self, constraint):
 
   def contains_path_match(self, path, match_spec, min=1, max=None,
                           **kwargs):
+    match_kwargs = kwargs.pop('match_kwargs', {})
     if isinstance(match_spec, list):
```
jtk54: If `match_spec` is neither a list or a dict, this will fail inside the dict branch with some nonsensical exception. Can you make the failure where the `match_spec` is neither a list or dict more explicit?

citest/json_contract/value_observation_verifier.py
```diff
@@ -80,22 +80,25 @@ def add_constraint(self, constraint):
 
   def contains_path_match(self, path, match_spec, min=1, max=None,
                           **kwargs):
+    match_kwargs = kwargs.pop('match_kwargs', {})
```
jtk54: It's not clear what the options for `match_kwargs` are without drilling into the predicates themselves. If the options are the same for each type of predicate, then I think a comment or docstring would be helpful for these functions.

citest/json_contract/value_observation_verifier.py
ewiseblatt: ok

citest/json_contract/value_observation_verifier.py
ewiseblatt: added class comment

PR #242, opened 2017-01-06
Modified gcp resource snapshot to accomodate general resource cleanup.

ewiseblatt:
@jtk54 
Turns out the delete_resource.py script I tried adding into spinnaker yesterday wasnt generalized enough to handle the container API. I'm adding a deletion (and enhanced filtering) capability into the resource snapshotter which has the sophistication I need to operate on arbitrary resources.


citest/gcp_testing/resource_snapshot.py
```diff
@@ -615,6 +699,55 @@ def __determine_added_instances(self, resource, before, after):
...
+      print '*** Cannot find delete method for "{0}"'.format(resource_type)
+      return None
+
+    scope = self.__explorer.pick_scope(
+        delete.get('scopes', []), api, mutable=True)
+    agent = self.make_agent(api, version, scope,
+                            default_variables=bindings)
+
+    decorator = '[dry run] ' if self.__options.dry_run else ''
+    print '{decorator}DELETING from API={api} with scope={scope}'.format(
```
jtk54: Is `resource_type` also relevant to include in this message?

ewiseblatt: This is the API and scope being used.
The individual elements show the type (and name and parameters)

PR #243, opened 2017-01-07
Improved upon resource listing.

ewiseblatt:
@jtk54


citest/gcp_testing/resource_snapshot.py
```diff
@@ -623,13 +651,78 @@ def transform(items):
...
+        'list', resource, **{param_name: 'tbd'})
+    items = []
+    for param_value in param_variants:
+      results = agent.list_resource(
+          context, resource, method_variant='list',
+          **{param_name: param_value})
+      items.extend([('{0}s/{1}'.format(param_name, param_value), item)
+                    for item in results])
+
+    return params, 'aggregatedList', items
```
jtk54: I'd wrap this tuple in an object. It might seem superfluous, but unnamed tuples aren't maintainable. With a class, the code will document itself.

PR #247, opened 2017-03-05
Memory

ewiseblatt:
@jtk54 or @lwander


ewiseblatt:
Our builds started failing OOM on the kato test after an influx of images
were added into the project.

PR #248, opened 2017-03-14
add os_agent & os_contract

ashleykasim:


Roshan2017:
I signed it!

ewiseblatt:
It turns out the error thing is ok. I havent reviewed this again since you made your changes. I'll get to it by morning.

ewiseblatt:
I am going to merge this. The github check error seems to be a problem on our end because of the multiple authors.

I do have a suggestion while testing your agent against openstack. You might want to check for some of the environment variables that you expect are set and raise an explicit error if they are missing. But I think only if in fact they are required. For example I forgot to set them and got an error back from openstack cli
ValueError: Missing value auth-url required for auth plugin password.

We can fix this in another PR since it isnt clear what, if any, checks to perform.

Roshan2017:
Okay!

ewiseblatt:
Actually, before I merge this, can you squash it?

And also run pylint on it and fix the formatting errors. You can address the others if you want if the suggestions make sense. I'm not strict about it. I'm not sure why the missing method errors are there. It's ok. I think it has something to do with assumptions in the disjunction where the object is created in the base class. It isnt a problem in practice.

pylint --rcfile=pylint.rc citest/openstack_testing/os_agent.py citest/openstack_testing/os_contract.py

ashleykasim:
@ewiseblatt squashed it, should be ready to go.

ewiseblatt:
Does this need python-openstackclient in requirements.txt?
It doesnt look like a static dependency to parse, but you need it for openstack to run, right?

ewiseblatt:
or do you want to leave it out since there are other client options?

ewiseblatt:
if you leave it out, add something into the readme about how to obtain a client.

Roshan2017:
I have added that in README at spinnaker/testing/citest under "Install Platform Dependencies" section

ewiseblatt:
did you push the update (squash first) since I'm ready to merge.

Roshan2017:
@ashleykasim has squashed and pushed the citest changes and I have squashed and pushed spinnaker changes. Also, created a PR https://github.com/spinnaker/spinnaker/pull/1496

ewiseblatt:
Thanks for doing this.

PR #256, opened 2017-04-06
feat(aws): python agent implementation

ewiseblatt:
This is just a simple API that wraps the AWS boto client to log/journal,
but provides a better direction than using the aws cli.

This deprecated AwsAgent and AwsContractBuilder in favor of more explicit
AwsCliAgent and AwsCliContractBuilder but keeps the old names around for now.

@jtk54


citest/aws_testing/aws_python_agent.py
```diff
@@ -0,0 +1,168 @@
...
+class AwsJsonEncoder(json.JSONEncoder):
+  def default(self, obj):
+    if isinstance(obj, datetime.datetime):
+      return obj.isoformat()
+    return json.JSONEncoder.default(self, obj)
+
+
+
+class AwsPythonAgent(PythonAgent):
+  """A service_testing.CliAgent that uses the aws tool to interact with AWS.
```
jtk54: I thought this wasn't a `CliAgent`.

citest/aws_testing/aws_python_agent.py
```diff
@@ -0,0 +1,168 @@
...
+    _response_field = kwargs.get('_response_field', None)
+    response = super(AwsPythonAgent, self).call_method(
+        _method, context, *pos_args, **kwargs)
+    return response if _response_field is None else response[_response_field]
+
+  def call_method_and_extract_singular_response(
+      self, context, _method, _response_field, *pos_args, **kwargs):
+    _exactly_one = kwargs.get('_exactly_one', True)
+    if '_exactly_one' in kwargs:
+      del kwargs['_exactly_one']
```
jtk54: Why clear this from the `kwargs`?

PR #257, opened 2017-04-11
Initial commit of azure agent/contract

rguthriemsft:
@ewiseblatt PTAL.  If good I can rebase and submit.  This is initial PR for citest Azure support

### DIFF
--- a/citest/azure_testing/az_contract.py
+++ b/citest/azure_testing/az_contract.py
@@ -1,0 +1,135 @@
+# Create the base agent for Azure CITest scenario
+
+"""Support for specifying citest.json_contract.Contract on Azure resources."""
+
+# Python modules
+import json
+import logging
+import traceback
+
+# CITest Modules
+from .. import json_contract as jc
+from ..json_predicate import JsonError
+from ..service_testing import cli_agent
+
+
+class AzObjectObserver(jc.ObjectObserver):
+  """ Observe Az resources"""
+
+  def __init__(self, az, args, filter=None):
+    """Construct the observer.
+
+    Attributes:
+        az = AzCloudAgent instance to use.
+        args: Commang-line arguments list to execute.
+        filter: If provided, then use this to filter observations.
+    """
+
+    super(AzObjectObserver, self).__init__(filter)
+    self.__az = az
+    self.__args = args
+
+  def __str__(self):
+    return 'AzObjectObserver({0})'.format(self.__args)
+
+  def collect_observation(self, context, observation, trace=True):
+    args = context.eval(self.__args)
+    az_response = self.__az.run(args, trace=trace)
+    if not az_response.ok():
+        observation.add_error(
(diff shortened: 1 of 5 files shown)


citest/azure_testing/az_contract.py
```diff
@@ -0,0 +1,135 @@
...
+          error_regex='(?:.* operation: Cannot find .*)|(?:.*\(.*NotFound\).*)')
+      disjunction_builder = jc.ObservationVerifierBuilder(
+          'Collect {0} or Not Found'.format(command))
+      disjunction_builder.append_verifier(error_verifier)
+
+      collect_builder = jc.ValueObservationVerifierBuilder(
+          'Collect {0}'.format(command), strict=self.__strict)
+      disjunction_builder.append_verifier_builder(
+          collect_builder, new_term=True)
+#      self.verifier_builder.append_verifier_builder(
```
ewiseblatt: You want this. What this does is make this clause's verifier look like this:

disjunction_builder
|
+----- OR -------+
|  ......................... |
error_verifier ...  collect_builder

When you later return the collect_builder, the caller will add more constraints onto that. The effect is that when the clause validates itself, it will check for the error or all the constraints that the caller adds later.
When you leave this out, you arent assigning any verifier at all. The caller will be adding constraints to the result, but there is no association at all and the disjuction _builder (with the error verifier) is not referenced anywhere.

citest/azure_testing/az_contract.py
```diff
@@ -0,0 +1,135 @@
...
+#          disjunction_builder, new_term=True)
+    else:
+      collect_builder = jc.ValueObservationVerifierBuilder(
+          'Collect {0}'.format(command), strict=self.__strict)
+#      self.verifier_builder.append_verifier_builder(collect_builder)
```
ewiseblatt: You want this to be in the else clause so that the constraints that the caller will add onto the return result will become the constraints on the ContractClause.

citest/azure_testing/az_contract.py
```diff
@@ -0,0 +1,135 @@
...
+            cli_agent.CliAgentRunError(self.__az, az_response))
+        return []
+
+    decoder = json.JSONDecoder()
+    try:
+      doc = decoder.decode(az_response.output)
+      if not isinstance(doc, list):
+        doc = [doc]
+      observation.add_all_objects(doc)
+      #self.filter_all_objects_to_observation(context, doc, observation)
```
ewiseblatt: I havent been using filters in practice, but if you were to use a filter then you'd need this. The idea of the filters was to cut down on observed objects processed (and reported) downstream.

rguthriemsft:
I have made the suggested edits.  let me know if there is anything else.

ewiseblatt:
LGTM.
Is the Google Contributor License Agreement (CLA) signed?

rguthriemsft:
Yes, I signed yesterday.

PR #261, opened 2017-04-15
feat(validation): Simplified ValueObservationVerifier API with AND/OR…

ewiseblatt:
@jtk54 
@rguthriemsft


citest/json_contract/observation_verifier.py
```diff
@@ -381,19 +381,75 @@ def export_to_json_snapshot(self, snapshot, entity):
...
+    This starts the first verifier clause. Additional verifiers can be added
+    later by using either AND or OR depending on how to combine them.
+
+    Args:
+       verifier: [ObservationVerifier]
+    Returns:
+       self
+    """
+    if self.__current_builder_conjunction:
+      raise ValueError('There are already verifiers added. Use OR or AND')
```
jtk54: Maybe add here that the `EXPECT` keyword was already used: "There is already one EXPECT verifier added. Use OR or AND.".

ewiseblatt: expect isnt technically required. The API still has other ways to acquire verifiers.
Requiring EXPECT might be a good idea though. I'll have to rewrite all the tests since it is new.

PR #270, opened 2017-06-21
chore(observation_verifier): Deprecated ValueObservationVerifier

ewiseblatt:
This is a major refactoring because ValueObservationVerifier was the
basis of verifying observations. The problem is that verifying errors
or observations as a whole was awkward.

Before a ValueObservationVerifier was a specialized ObservationVerifier
that applied a predicate to the list of observed values. Since this was
the values and not the list, you couldnt see the obesrvation as a whole.
To get around this you coudl add a "constraint" which was an ObservationVerifier
that could look at the whole observation.

Also, there was a special verifier, ObservationFailureVerifier, which could
be used as a constraint to check the observation errors where you could
override a method "_error_comment_or_none" to confirm the error with a string
or return None to fail. This interface is non-stadnard and doesnt fit with the
rest of the toolkit.

The new API gets rid of ValueObservationVerifier entirely and introduces
a ObservationPredicate (along with specializations ValueObservationPredicate and
ErrorObservationPredicate). An ObservationPredicate is a ValuePredicate but
expects the value to be an Observation. The Value and Error ObservationPredicates
each take a standard ValuePredicate and apply it to the observation.objects or
observation.errors respectively. Now to check values or errors, you simply
add the observation predicates to the verifier and they are treated more
cohesively. The verifier still uses disjunctive normal form (DNF) so has
a list of lists of predicates rather than list of lists of verifiers.

It keeps the ValueObservationVerifierBuilder for the syntactic sugar it
provided for constructing verifiers. However the builder builds an
ObservationVerifier instead of a ValueObservationVerifier.
The ValueObservationVerifier no longer exists at all. Because
all consumers used the Builder in practice, they should not be affected
by this class going away entirely. Only the unit tests used this class
directly; they now use the builder to create the verifier instead.

Because this is a major API change, I deprecated the old interface in
such a way that it actively converts into the new one. This means that
all existing code in citest and spinnaker repositories still work
even though they dont use the new model yet. They will warn to logging
about this and future PRs can transform and cleanup the rest of the code base
incrementally in more manageable sized PRs.

@jtk54 
@rguthriemsft FYI


citest/json_contract/observation_predicate.py
```diff
@@ -0,0 +1,224 @@
...
+  expected to be a json_contract.Observation.
+  """
+  pass
+
+
+class ObservationErrorPredicate(ObservationPredicate):
+  """An observation predicate that mediates Observation.errors.
+
+  Note that this mediates individual errors,
+  not the whole list of errors.
```
jtk54: Why is this class constrained to not be able to operate on the whole list, like `ObservationValuePredicate`?

citest/json_contract/observation_verifier.py
```diff
@@ -314,20 +344,25 @@ def __call__(self, context, observation):
...
     # Outer terms are or'd together.
     for term in self.__dnf_verifiers:
        # pylint: disable=bad-indentation
        term_valid = True
        # Inner terms are and'd together.
        for v in term:
           result = v(context, observation)
-          builder.add_observation_verify_result(result)
+          if isinstance(result, ObservationVerifyResult):
+            logging.info('Deprecated embedded ObservationVerifyResult')
```
jtk54: Should this be a warning?

PR #271, opened 2017-06-26
breaking change: GcpTesting

ewiseblatt:
This converts over the gcp_testing to the new API,
especially handling observation errors. These are no longer
options into the observer method, rather they are specified
by the caller's ObservationVerifier (such as the
new ObservationErrorVerifier).

I converted over the spinnaker tests in another PR. I'm not aware
of other use cases that tested for errors the way removed so this
so am removing the obsolete implementation here and now as well
making this a breaking change.

This PR also adds a factory class for creating predicates to complement
ObservationVerifier EXPECT/OR/AND methods so as to keep that
api more lean. This is still more unweildy than I'd like, but
is more consistent and clearer than the original ValueObservationVerifier
API was.

I could add more specialized constructor varianets but am holding off
on that. At least now it is more clear and consistent to use.

@jtk54


jtk54:
Please you change your commit message to `feat()` and add a line in the commit message containing `BREAKING CHANGE` (caps included) to signal a breaking change.

citest/gcp_testing/gcp_error_predicates.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, http_code=None, content_regex=None, **kwargs):
     self.__content_regex = content_regex
     super(HttpErrorPredicate, self).__init__(**kwargs)
 
-  def __call__(self, value):
+  def __call__(self, context, value):
```
jtk54: Is this just to match the interface? Doesn't look like it's used.

citest/json_contract/observation_verifier.py
```diff
@@ -446,122 +447,78 @@ def export_to_json_snapshot(self, snapshot, entity):
...
+       pred: [ObservationPredicate]
     Returns:
        self
     """
     if self.__current_builder_conjunction:
       raise ValueError(
-          'EXPECT (or append_verifier) was already used.'
-          ' Use OR or AND to add more verifiers.')
+          'expect (or "AND()" or "OR()") was already used.'
+          ' Use "OR()" or "AND()" to add more predicates.')
```
jtk54: This means there's already a conjunction started, right? Seems like the first parenthesis `(or "AND()" or "OR()")` is a bit confusing here. Maybe something like `Conjunction already starting using `expect`, use `OR()` or `AND()` to add more predicates.

tests/gcp_testing/gcp_contract_test.py
```diff
@@ -92,9 +93,9 @@ def test_inspect_not_found_ok(self):
     contract_builder = gt.GcpContractBuilder(agent)
 
     c1 = contract_builder.new_clause_builder('TITLE')
-    verifier = c1.inspect_resource(
-        'regions', resource_id='us-central1-f', no_resource_ok=True)
-    verifier.contains_path_value('field', 'value')
+    verifier = c1.inspect_resource('regions', resource_id='us-central1-f')
+    verifier.EXPECT(
+        ov_factory.error_list_contains(gt.HttpErrorPredicate(http_code=404)))
```
jtk54: Nice.

citest/gcp_testing/gcp_error_predicates.py
ewiseblatt: yes. ValuePredicates always take a context, value as the interface.
When I added context in I must have overlooked this class somehow.
This method will be used when you bind this predicate to an ObservationErrorPredicate to delegate error analysis to.

ewiseblatt:
one last look...

PR #274, opened 2017-07-07
chore(refactor): Refactored binary_predicate module.

ewiseblatt:
I broke apart the binary_predicate module into a few submodules
to decouple the internal classes. I left the compound predicates
that have circular references as they decompose instances in place.
The others were broken out only for the purposes of reducing scope
of the individual modules.

I renamed "StandardBinaryPredicate" to "SimpleBinaryPredicate"
but preserved the old class name for now, issuing warnings on use
so this is not a breaking change. It is no longer referenced internally.

There are no functional code changes. This is only moving code into refined
modules (and breaking existing tests into corresponding modules), and
changing affected imports.


ewiseblatt:
@jtk54

PR #275, opened 2017-07-07
feat(predicate): Added STR_REGEX

ewiseblatt:
@jtk54


citest/json_predicate/simple_binary_predicate.py
```diff
@@ -104,6 +105,11 @@ def __call__(self, operand):
 STR_NE = SimpleBinaryPredicateFactory(
     '!=', lambda a, b: a != b, operand_type=basestring)
 
+STR_REGEX = SimpleBinaryPredicateFactory(
+    'RegEx',
+    lambda a, b: re.match('^{0}$'.format(b), a) != None,
```
jtk54: Are the `^` and `$` necessary here? Seems like this would muddle the interface a bit. Removing would match the interface of `re.match` in the predicate and allow it to be more flexible.

ewiseblatt: I changed match to search and removed these.

PR #286, opened 2018-05-09
fix(base_test_case): Handle generic TestResult for other runners

ewiseblatt:
(e.g. py.test)

@jtk54 
I just found this in my client that I never sent out but have been using.
I've lost the train of thought as to whether or not it is complete, but
it is at a milestone where you can run pytest on the all the unit tests where
before you could not on those tests using the AgentTestCase, which is the
basis of most citests.


citest/base/base_test_case.py
```diff
@@ -123,41 +126,63 @@ def __end_step_context(self, relation):
...
+
     if result is None:
       result = self.defaultTestResult()
 
-    num_failures = len(result.failures)
-    num_errors = len(result.errors)
+    do_skip = result.addSkip
+    do_success = result.addSuccess
+    do_error = result.addError
+    do_failure = result.addError
```
jtk54: Do we want `addError` or `addFailure` in this case?

citest/base/base_test_case.py
```diff
@@ -123,41 +126,63 @@ def __end_step_context(self, relation):
...
+    return delegate(test)
+
+  def __trap_failure(self, delegate, test, err):
+    self.__final_outcome_relation = 'INVALID'
+    return delegate(test, err)
+
+  def __trap_error(self, delegate, test, err):
+    self.__final_outcome_relation = 'ERROR'
+    error_details = '%s: %s' % (err[0], err[1])
+    JournalLogger.journal_or_log_detail(
```
jtk54: What's the reasoning for logging on error only? Would we want to log on failure as well?

citest/base/base_test_case.py
ewiseblatt: Good catch.

citest/base/base_test_case.py
ewiseblatt: I think failures would already be logged because of the reporting built into the framework.
The errors are more likely to occur outside the framework thus not be logged.

PR #287, opened 2018-05-09
feat(HttpAgent): Added HttpResponsePredicate

ewiseblatt:
@jtk54


citest/service_testing/http_predicate.py
```diff
@@ -0,0 +1,127 @@
...
+    self.__value = value
+
+  def __eq__(self, result):
+    return (super(HttpResponsePredicateResult, self).__eq__(result)
+            and self.__value == result.value)
+
+  def export_to_json_snapshot(self, snapshot, entity):
+    """Implements JsonSnapshotableEntity interface."""
+    builder = snapshot.edge_builder
+    builder.make_output(entity, 'HTTP Code', self.__value.http_code)
```
jtk54: Is there a request path we can add to this?

ewiseblatt: This typically represents the value that was compared with the predicate, not its pedigree.
If the source were of interest, then it would probably also be elsewhere in the snapshot context.
For example if this is an observed value, the observation request would be in the snapshot.
The path would be there.

PR #292, opened 2018-07-06
fix(agent_test_case): Fix undefined variable and pass through run_tes…

ewiseblatt:
…t_case_list params

@jtk54


ewiseblatt:
@jtk TBR

citest/service_testing/agent_test_case.py
```diff
@@ -80,10 +80,9 @@ def default_relation(self):
     """
     if self.__exception is not None:
       return 'ERROR'
-    elif self.__verification is not None:
+    if self.__verification is not None:
```
jtk54: Is `__verification` a boolean here or something different? Seems like the ternary in the return should use the same condition unless I'm missing something.

ewiseblatt: I did this as a result of a lint warning. Looking into your comment it seems that __verification (and __verification_summary) is never set anywhere. I'm going to leave this change since it is equivalent to what was before and look into it later. __verfication would have been the object [SnapshotableEntity] representing the validation results.

I'm not sure what you mean by "use the same condition". What is the reference point? The logic here is the same as before. The difference is removing the else clauses because the true case would have returned already so the else was deemed superfluous.

jtk54: SGTM.

PR #294, opened 2018-07-19
feat(gcp): inspect api resources.

ewiseblatt:
@jtk54 

This PR is a refactoring of the gcp_testing/resource_snapshot.py program into some modules that can be callable directly outside the original specialized CLI.


citest/gcp_testing/api_investigator.py
```diff
@@ -0,0 +1,376 @@
...
+        continue
+      catalog[entry['name']] = entry['version']
+
+    ApiInvestigator.__api_version_map = catalog
+    return catalog
+
+  @staticmethod
+  def get_api_version(api):
+    """Return current version of requested api name."""
+    return ApiInvestigator.collect_apis()[api]
```
jtk54: Can we collect the APIs from the discovery doc once instead of for each API call?

citest/gcp_testing/api_investigator.py
```diff
@@ -0,0 +1,376 @@
...
+  def get_api_resource_filter(self, api):
+    """Returns ApiResourceFilter for named api, or None if not known."""
+    return self.__api_to_resource_filter_map.get(api)
+
+  def get_listable_unlistable_scope_map(self, api):
+    """Determine the listable and unlistable methods and the scope to use.
+
+    Typically the individual accessor methods are used instead of this.
+    Returns:
+       (listable, unlistable, scope_map) for the given API
```
jtk54: Wrap in a named tuple for readability?

citest/gcp_testing/api_investigator.py
ewiseblatt: collect_apis() is on demand, then caches for future calls. So it already is once as you suggest.

citest/gcp_testing/api_investigator.py
```diff
@@ -0,0 +1,376 @@
...
+
+from citest.gcp_testing.gcp_agent import GcpAgent
+from citest.base import ExecutionContext
+
+
+class ApiResourceFilter(object):
+  """Determines which resources in an API are wanted."""
+  # pylint: disable=too-few-public-methods
+
+  def __init__(self, include_regexs, exclude_regexs):
```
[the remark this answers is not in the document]
ewiseblatt: changed to regex_list

PR #296, opened 2018-07-24
chore(python3): Port to python 3

ewiseblatt:
@jtk54 
This is an initial attempt to port to python 3 (while retaining compatability with python 2).


citest/base/bindings.py
```diff
@@ -20,8 +20,16 @@
...
 import os
 import re
+import sys
+
+try:
+  import ConfigParser
+except ImportError:
+  import configparser as ConfigParser
+
+if sys.version_info[0] > 2:
```
jtk54: Is this either 2 or 3? Will `version_info` ever contain a minor, e.g. 2.7?

citest/base/global_journal.py
```diff
@@ -58,8 +58,8 @@ def new_global_journal_with_path(path, **metadata):
       atexit.register(_atexit_handler)
       _added_atexit = True
 
-    journal_file = open(path, 'w')
-    os.fchmod(journal_file.fileno(), 0600)  # Protect sensitive data.
+    journal_file = open(path, 'wb')
+    os.fchmod(journal_file.fileno(), 0o600)  # Protect sensitive data.
```
jtk54: Should we use `stat` constants instead of the raw file permissions? https://stackoverflow.com/questions/16249440/changing-file-permission-in-python

citest/json_predicate/binary_predicate.py
```diff
@@ -177,6 +182,38 @@ def _verify_elem(self, context, elem, the_list):
     return False
 
 
+class ListSimilarPredicate(_BaseListMembershipPredicate):
```
jtk54: Relevant to the port?

citest/json_predicate/binary_predicate.py
```diff
@@ -360,25 +397,14 @@ def __call__(self, context, value):
...
-  sorted_a = sorted(a)
-  sorted_b = sorted(b)
-  for index, value in enumerate(sorted_a):
-    if sorted_b[index] != value:
-      return False
-  return True
-
-LIST_SIMILAR = SimpleBinaryPredicateFactory(
-    '~=', lambda a, b: lists_equivalent(a, b), operand_type=list)
+#LIST_SIMILAR = SimpleBinaryPredicateFactory(
```
jtk54: Remove.

citest/json_predicate/binary_predicate.py
ewiseblatt: actually yes, but probably should have done that in a separate PR.
In python 3 you cannot compare dictionaries. We used to use that before to compare list membership by first sorting to it would be easy to compare. Now I cannot figure out how to do something efficient so I'm using the subset check both ways to test a <=b and b <=a because the subset check has logic in it.

ewiseblatt: I have it in this PR because it is tied to python3 since that's the only reason for the change.

PR #307, opened 2018-09-25
feat(reporting): Added CLI for manipulating journals

ewiseblatt:
This adds a simple script for creating journals and appending messages
into them, particularly to create reports for tests that were unable
to execute for whatever reason.

@zimanyi


citest/reporting/html_index_renderer.py
```diff
@@ -119,7 +119,7 @@ def __handle_generic(self, entry):
     if entry.get('control') == 'END':
         # pylint: disable=bad-indentation
         self.__depth -= 1
-        if self.__depth == 0 and self.__in_test:
+        if self.__depth == 0: # and self.__in_test:
```
ezimanyi: Was this change deliberate?

citest/reporting/journal_cli.py
```diff
@@ -0,0 +1,328 @@
...
+   2) To create journal files to indicate that tests could not be run.
+      Normally journal files are created while running tests. However if
+      the tests cannot be run at all (e.g. precondition failed in a launcher)
+      then there is no way to indicate that through citest output since citest
+      wont even be called.
+"""
+
+import argparse
+import json
+import logging
```
ezimanyi: Looks unused

citest/reporting/journal_cli.py
```diff
@@ -0,0 +1,328 @@
...
+
+  options = parser.parse_args(argv)
+  command = commands.get(options.command, not_found)
+  command(options)
+  sys.exit(-1 if command == not_found else 0)
+
+
+def init_logging(options):
+  """Initialize logging within this tool."""
+#  log_file = options['log_basename'] + '.log'
```
ezimanyi: Should these lines be uncommented?

citest/reporting/journal_cli.py
```diff
@@ -0,0 +1,328 @@
...
+         'handlers':['console']
+       },
+    }
+  }
+  if log_dir:
+    log_config['handlers']['file'] = {
+      'level':'DEBUG',
+      'class':'logging.FileHandler',
+      'formatter':'timestamped',
+      'filename': os.path.join(log_dir, log_file),
```
ezimanyi: Missing `import os`

citest/reporting/journal_cli.py
ewiseblatt: I didnt mean to push this change.

citest/reporting/html_index_renderer.py
```diff
@@ -119,7 +119,7 @@ def __handle_generic(self, entry):
     if entry.get('control') == 'END':
         # pylint: disable=bad-indentation
         self.__depth -= 1
-        if self.__depth == 0 and self.__in_test:
+        if self.__depth == 0:
```
ezimanyi: I'm not sure I understand why the `and self.__in_test` was removed.  The first commit commented it, and the second removed it entirely (whereas I was expecting the second commit to just uncomment)

ewiseblatt: The clause used to be there because I only wanted to consider test failures.

When I added the CLI, it is pushing journals that arent running tests because the whole point is that it is reporting that it could not initiate the test. I commented out that clause to see if I needed it before I yanked it, but forgot to yank it.

Note that when I added the earlier change to catch issues in the setup/teardown which is also outside the tests, that was pretty much having the same effect of annotating top-level entities that arent in tests (but was outside contexts as well).

The net effect here is that rending journals as HTML isnt really about tests. It's really just a structured log file, which was the intent all along. But it isnt practical to think of it completely generalized because it isnt efficient and has limited expressiveness. It's still specialized for citest needs.

ezimanyi: Thanks for the explanation!  This makes perfect sense now.

PR #313, opened 2018-12-11
feat(reporting): Extract timing statistics from journals.

ewiseblatt:
@jtk54


ewiseblatt:
This is to separate performance from functional testing in the tests. We've
been loosening timeouts because of jitter and flakiness across various
internal and external services. The intent was to let a test go into excess
to pass, then understand what it took and decide by some other policy
outside the individual tests.

PR #315, opened 2018-12-21
feat(html): Support format=yaml

ewiseblatt:
@jtk54


citest/reporting/html_renderer.py
```diff
@@ -210,6 +217,47 @@ def process_json_html_if_possible(self, value):
...
+    """
+    summary = ''
+    document_manager = self.__document_manager
+    to_yaml = lambda value: yaml.safe_dump(
+        value, indent=2, default_flow_style=False)
+    try:
+      if isinstance(value, basestring):
+        doc = yaml.safe_load(StringIO(value))
+        text = to_yaml(doc)
+        if text.endswith('...\n'):
```
jtk54: Which delimiter is this?

ewiseblatt: It turns out that if you have a YAML file which is just a string "hello" then the YAML document looks like "hello\n...\n"
https://yaml.org/spec/1.1/#id857577

So in the process of trying to determine if the string is a YAML document or just a string, the conversion into a YAML document winds up producing one that is formally terminated with "...", which is not what we intended to render here.

jtk54: Huh. May be worth pointing that part of the spec out here in a comment.

PR #321, opened 2019-03-20
fix(ssl): Initialize self.__ignore_ssl_cert_verification in HttpAgent's constructor.

joonlim:


joonlim:
Merging preemptively to fix internal issue we are having.

PR #325, opened 2019-05-07
Add more logging for long deletes.

plumpy:
Log the resource type that we're waiting for. If we give up waiting for
certain resources to be deleted, list those resources.

Also bumped up the timeout for deletion from 3 to 5 minutes.

And fixed --print_api_spec.


plumpy:
cc @ewiseblatt

PR #326, opened 2020-08-12
fix(python3): remove async reserved word

allisaurus:
In python 3.7 and later, ["async" is a reserved word](https://docs.python.org/3/whatsnew/3.7.html). This change renames a variable with that name to avoid syntax errors when running with these later versions of Python.

Also adds a check re: the presence of the [`os.fchmod`](https://docs.python.org/3/library/os.html#os.fchmod) attribute to avoid errors when running on Windows.

### DIFF
--- a/citest/base/global_journal.py
+++ b/citest/base/global_journal.py
@@ -61,7 +61,8 @@ def new_global_journal_with_path(path, **metadata):
 
     journal_file = open(path, 'wb')
     # Protect sensitive data.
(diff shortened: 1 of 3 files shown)


allisaurus:
FYI: per [CONTRIBUTING](https://github.com/google/citest/blob/master/CONTRIBUTING), I should be authorized to contribute under the Amazon CLA. LMK if you need more info here.

cc @plumpy who I chatted with [on slack](https://spinnakerteam.slack.com/archives/CKXFWFB53/p1595541082323500) about this.

plumpy:
Thanks, Allie!

